PR #415, opened 2014-12-09
Add CERT IDENTIFY command to NickServ.

AppleDash:
This adds a new subcommand to the CERT command which allows you to re-identify to NickServ using your current CertFP, without reconnecting.


kaniini:
@ircv3 merged this and atheme supports it out of the box so all good

PR #432, opened 2015-01-14
Fix GroupServ issues

shockkolate:
This pull request eliminates issues arising from hardcoded parsing of flags in the GroupServ flags parser. The GroupServ FLAGS help text is updated to reflect the now functional GroupServ flag +A (allows viewing of group access list). Additionally, issues with flags +* are fixed and now the behaviour is more consistent with that of ChanServ.

More details are available in the individual commit messages.


Renegade334:
To get it closer to ChanServ, there is a pre-existing quirk that should probably be addressed: if a FLAGS command is issued on a user who is not a member of the target group, and the flags string provided results in no flags being allocated to the user (eg. `FLAGS !test nick +XYZ` or `FLAGS !test nick -*`) then the user will be added to the group but with no flags. Performing the same command a second time will remove them from the group for having no "remaining" flags.

ChanServ's FLAGS performs a check on the flagstring and fails the command if there are no valid flags. In addition, it will never leave a chanacs line with no flags.

Renegade334:
@shockkolate: the same improvements/fixes should also be applied to `fflags.c`.

PR #434, opened 2015-01-17
Implement better PBKDF2 hashing for passwords

aaronmdjones:
This version embeds the PRF and iteration count into the crypt()-like
strings it generates, meaning that you can change these defaults in
the module at a later date without invalidating all of your current
hashes. You can also extend it with new PRFs for this reason.

The usage of sscanf(3) in this module is consistent with ISO C89 and
there is a comment in the source file reminding people not to change
the format strings carelessly.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -88,6 +88,11 @@ other
 - contrib/cap_sasl.pl: Fix crash if disconnected while waiting for SASL reply
 - transport/jsonrpc: new module implementing JSONRPC transport
 
+crypto
+------
+- pbkdf2v2: Newer module implementing PBKDF2-HMAC digest scheme
+            with backward compatibility and limited forward compatibility
+
 Atheme Services 7.1 Release Notes
 =================================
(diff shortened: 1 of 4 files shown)


jillest:
The within-provider compatibility does not work as well as compatibility between different providers (crypto modules). There is code in verify_password() to upgrade passwords from older providers to the default provider, but this cannot upgrade passwords from older parameters to the parameters for new passwords. Such upgrading is then only possible by manual user SET PASSWORD.

aaronmdjones:
Oh, no, I am aware. The comment reflects the fact that if the module is upgraded in the future, it can continue to validate all previous hashes it generated. That is not the case for the current module.

kaniini:
As previously mentioned I'm not a huge fan of those `sscanf()` calls.

jillest:
The `sscanf()` calls look correct to me, but I haven't tested it.

aaronmdjones:
I have tested it, extensively, in production even. Linux, amd64, GNU libc 2.19

PR #435, opened 2015-01-18
protocol: Fail extban check if mask contains space

shockkolate:
In certain circumstances, the mask string passed into is_extban can contain space(s) and currently there is no validation in any protocol module implementing extbans that the input mask indeed contains no spaces. An example circumstance is ChanServ's BAN command.

The following command will cause ChanServ to set a ban on the first mask due to how the MODE command works, but fail to add the ban to Atheme itself due to the mask being the full rest of the string as opposed to just a single word.

`/msg ChanServ BAN #channel $a:mask1 $a:mask2`

This patch adds a simple strchr check for the space character to each extban check, fixing the issue outlined above.

### DIFF
--- a/modules/protocol/charybdis.c
+++ b/modules/protocol/charybdis.c
@@ -291,6 +291,9 @@ static bool charybdis_is_extban(const char *mask)
 	if ((mask_len < 2 || mask[0] != '$'))
 		return NULL;
 
+	if (strchr(mask, ' '))
+		return false;
+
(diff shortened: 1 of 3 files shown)


shockkolate:
Note: This patch has been tested on Charybdis and InspIRCd but has not been tested on UnrealIRCd. The change for each case is basically exactly the same but it should probably be tested for just-in-case sake.

PR #436, opened 2015-01-18
Implement a within-module parameter upgrade system to satisfy

aaronmdjones:
some concerns about parameter upgrades over time

Also tidy up some styling and add a comment about salt length.

NOTE: Other modules need to be investigated for possible integration
      into this scheme


kaniini:
Looks fine except for a few style nitpicks I pointed out inline.

PR #437, opened 2015-01-18
Add +e and change "SaslServ"

Spydar007:
Add +e to chanserv{} templates and founder_flags, and change SaslServ to SASLServ as per acronym capitalization rules.

### DIFF
--- a/dist/atheme.conf.example
+++ b/dist/atheme.conf.example
@@ -1177,11 +1177,11 @@ chanserv {
 	 */
 	templates {
 		vop = "+AV";
-		hop = "+AHhitrv";
-		aop = "+AOhiortv";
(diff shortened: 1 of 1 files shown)


ilbelkyr:
I feel like these changes should be in two separate commits; they seem quite unrelated.

jillest:
I like the templates part but not the saslserv renaming part.

Mikaela:
:+1: to ibelkyr and I don't think SaslServ should be renamed. It would become different from NickServ and other \*Servs and everyone is used to it being SaslServ, but this is just how I feel.

kaniini:
:-1: on the SaslServ renaming.  Patch rejected.

PR #448, opened 2015-02-03
saslserv: optionally call bad_password on SASL authentication failure

ilbelkyr:
As requested by @tomaw, this calls `bad_password` on users failing to identify to an account via SASL to help spot brute force attacks on passwords. Given that this might be a little confusing to users and doesn't include much info about the actual source, this can be toggled by a config option.


spb:
I don't think this should be optional.

PR #458, opened 2015-03-14
Fix some problems with unregistered target nicks in NickServ GHOST

Renegade334:
Fixes the following issues:

* The checks for frozen-ness and the nopassword flag were bypassed for unregistered targets (also fixes an assert fail from `metadata_find` when this occurred)
* If a password was provided, it was ignored for unregistered targets


Renegade334:
Retrospective addendum: 5d1ff1b36347ba88d75a518cbedd6941ba6dac01

Issue #397, opened 2014-11-02
Nick-/ChanServ: Registering "clear", "modify" or "list" breaks /msg ChanServ flags #channel clear/modify/list


ghost:
There is some kind of ironic "Anope compatibility" implemented in ChanServ here:
https://github.com/atheme/atheme/blob/master/modules/chanserv/flags.c#L195

It seems to be a feature, but it is completely weird, that any user can register the nicks "clear", "modify" and "list", resulting in a network-widely changed command syntax:
#### before registration:

"/msg ChanServ flags #channel list"
--> shows the flags for #channel
"/msg ChanServ flags #channel modify Username +-modification"
--> modifies the flags for Username on #channel
"/msg ChanServ flags #channel clear"
--> _clears the channel's flags_ (!)
#### after registration:

"/msg ChanServ flags #channel list"
--> shows the flags for the user who grouped the nick "list"
"/msg ChanServ flags #channel modify Username +-modification"
--> _[ChanServ] Invalid template name given, use /msg ChanServ TEMPLATE #channel for a list_ (!)
"/msg ChanServ flags #channel clear"
--> shows the flags for the user who grouped the nick "clear"
#### issues caused by this behavior:

This is highly theoretical, but it could actually lead to annoyance in certain cases:
- If Alice uses the nick "clear" and the founder of #channel wants to give them privileges, Alice could quickly drop/ungroup their "clear" nick to cause the founder to clear their whole #channel flags list. This can be _very_ annoying and hard to restore. As it - as far as I know - also includes all AKICK entries, it could even be abused to cause the founder to unban all people on the AKICK list.
- If there is someone who has a nick that matches a template, the FLAGS MODIFY command can be abused. Let's say someone added "lover" as a template for +AVeiorstv in their #channel (this actually happened). This someone is used to the Anope compatibility syntax and wants to give the user "LoVeR" the +V flag, and nothing more.

This someone now wants to use the following command to first check the current flags of LoVeR:
/msg ChanServ flags #channel modify LoVeR
--> Expected result: [ChanServ] No flags for LoVeR in #channel.
Then they would like to use "/msg ChanServ flags #channel modify LoVeR +V" to give them the needed flag.
However, LoVeR quickly and silently - for example using an identified webchat in the background - groups the "modify" nick to their account. Now someone uses the command:
/msg ChanServ flags #channel modify LoVeR
[...]

PR #472, opened 2016-03-05
Remove problematic Anope "compatibility". See comment

ToBeFree:
Moved to a comment because Atheme might not be using GitHub in the future.

Fixes GitHub issue atheme:#397


kaniini:
NAK.  Block registration of the keyword nicks instead.  This syntax was added for a reason.

ToBeFree:
Block registration of nicks that might well already be registered? Not a fix? :)

PR #512, opened 2016-10-03
Remove GameServ and RpgServ

maxteufel:
As GameServ and RpgServ just add more code to maintain that brings little to no benefits for users, we can remove them. This kind of functionality can and should rather be implemented using normal channel bots.


Xenthys:
They could probably be added to the contrib repository instead, so people can still find them if they really want those features in their services, but know it's not officially maintained as part of the core anymore.

siniStar7boy:
While I agree with @maxteufel, I also understand that this is going to probably cause some upset in some networks that actually use these..  Maybe moving them to contrib as @Xenthys mentioned _would_ be a better idea. _(no longer maintained, but still available for networks that feel like they cannot live without them.)_

Xenthys:
I must say I totally agree to see them disappear from the core, they are not useful for channel management nor anything that could be important. I suggested to move them to contrib because there are still things like LoveServ there, which is certainly also an unmaintained code.

Zarthus:
To give a short summary of what I'm about to say: I think GameServ sees more use than RPGServ, and RPGServ is a bit of an anti-service in some scenarios. If the goal is to remove services however, I propose it is done in a major version change, and not a minor version change (8.0.x versus 7.3.x). In addition, I've summarized my thoughts on each server at the end of each section.

----

I think it's worth it to discuss RPGServ and GameServ individually.

I will note that the below comment on RPGServ is not true for everyone, there are casual roleplaying communities that exist of (what some people would call) bad roleplay, but they're having a good time -- So I'm not one to judge.

As far as RPGServ goes, its features seem advanced - but I think to rely on services to find roleplay themed channels is a better job for ALIS than it is for a dedicated service. I don't know if many networks load RPGServ but based on the code it seems like it does roleplaying a disservice. With the stage IRC is at in this day and age it's more and more common for roleplaying channels to not want to be found, because the quality of a roleplaying channel decreases as you let people inside without checking if they are a good fit. I don't think RPGServ serves a good purpose if the end-goal is to help channels find good roleplaying channels.

Roleplaying communities tend to have a stricter process for finding people to roleplay with, usually thorugh forums. If networks are looking to ditch RPGServ while still maintaining its core functionality, a simpler answer is to load ALIS and topic search for roleplaying keywords, especially since relying on a predetermined (and unconfigurable) list is just bad (for roleplaying, and for maintaining).

With that said, I don't think it serves much as a purpose at all - but it's definitely seeing usage on dedicated roleplaying networks. If you want feedback from such a network, it might be an idea to see if the Darkmyst operators are willing to share how they feel about it.

If it wasn't already clear, I'm not opposed to the removal of RPGServ.

---

GameServ sees more frequent use on networks, and while I am unable to provide statistics on how many channels use it on EsperNet, it's safe to say that it sees occasional use. I won't mourn its loss if it were removed, we only use a very small subset of its functions (rolling, primarily).

GameServ can be replaced by many channel level bots, but not everyone has access to hosting them and some just don't want to go through that amount of effort. I think providing a rolling command is pretty nice, although I do agree dedicating an entire service to it is a bit over the top. I don't think its namegen function is particularily useful (There are a ton of websites which do this, better.)

The 8ball command is something you can find in every bot, and I personally believe most of the fun comes from supplying custom responses to that - which would be infeasible for Atheme. The direction I'm taking here is that this module is mostly redundant.

The Rock, Paper, Scissors is for playing RPS with an AI.  Fun.

The lottery command has some uses if you want unbiased giveaways or something, but todays IRC has a lot of idlers through bouncers and I think its function is a bit diminished. It's help page seems to indicate it's a fantasy-only-command too. And I'm personally not aware of any clients that don't support simple scripting like picking a name from the nicklist. Overall, might be nice to port to contrib, but I don't think this is hugely needed or used either.

Once you remove the namegen function you're left with mostly commands that roll dice for you in different manners, and a method to calculate your rolls for you. I think it's fair to conclude that all gameserv basically is is a "way to roll dice". And maybe it's an idea to just move that to a fantasy-enabled-command in contrib. All OSes ship with calculators by default and most of the roleplaying maths you do can be done in your head.

In conclusion, the main benefit of gameserv is that you can configure it to be only used by ops/voices/etc. But if you strip the service down to all useful commands it brings: You're left with a roll command. That's probably best placed in the contrib folder.

----

My final concern is the debate on whether this is a thing you want for Atheme 7.3, and not for 8.0.

This really depends on what you want to do, but I think if you're going to start removing components many people use, 7.3 is the wrong version to do it in.

kaniini:
*grumble*

staticfox:
Is there anything ground breaking that warrants removing those two services? I don't see very much activity in the repo, so it strikes me as odd that those two modules are being removed prior to any major refactors going on in core that would also make refactoring all other modules a requirement.

AFAICT it looks like most of the code within Atheme itself has not been touched, outside of minor things. So to drop things due to "maintainability" when not much else is going on in the project sounds like more of an excuse than a reason. If you are that concerned about exploits, why not audit it? In the grand scheme of things, it would most likely be faster to audit those modules then to discuss it in this pull request that seems out of the blue in the first place. (Unless I missed a discussion within an issue?)

> This kind of functionality can and should rather be implemented using normal channel bots.

I agree, however, it should most likely _at least_ be put in to contrib rather then dropping stuff that has been around for years.

kaniini:
Can someone actually explain the technical debt keeping these services incurs?  Because for the entirety of their existence so far, it's been basically zero.

kaniini:
Still waiting after 6 days...

siniStar7boy:
I tend to agree with @kaniini and @staticfox - there has not been any technical debt keeping these services.  Unless there is a code audit that proves there are indeed concerns or attack vectors, _I propose this PR and plans to remove the services - be closed/killed._

maxteufel:
Looks like I missed that.

However, I agree now that this should be closed.

PR #519, opened 2016-10-11
Add note about segfault on OpenBSD

williamleuschner:
This pull request adds documentation to the example configuration file that warns Atheme administrators of a potential segmentation fault when attempting to use the `modules/crypto/posix` module on OpenBSD.

I don't know how you feel about pulling to master, but I figured that since this is only a change to a comment, it wouldn't be _too_ heinous.

### DIFF
--- a/dist/atheme.conf.example
+++ b/dist/atheme.conf.example
@@ -105,6 +105,13 @@ loadmodule "modules/backend/opensex";
  * If you have several thousand accounts, this conversion may take
  * appreciable time.
  *
+ * A note to OpenBSD users: OpenBSD's crypt() implementation returns
+ * null when you tell it to do md5crypt, since MD5 is no longer
+ * considered safe for general use. As such, if you try to run Atheme
+ * with the modules/crypto/posix module, it will segfault as soon as it
+ * tries to hash something. Thankfully, OpenBSD includes LibreSSL by
+ * default, meaning that you can load PBKDF2 without issue.
+ *
  * The following crypto modules are available:
  *
  * PBKDF2 cryptography (new)                    modules/crypto/pbkdf2v2


grawity:
Hold on, why is this being "fixed" with a comment in example.conf rather than having crypto/posix actually check for failure? Surely OpenBSD being picky is not the only situation where crypt() can return null... Am I missing something?

dwfreed:
@grawity sure, that should be fixed too, but crypto/posix won't actually work at all on OpenBSD, because it falls back to DES, which is even worse.

williamleuschner:
I don't know C, so I won't be able to fix this at a deeper level.

grawity:
> it falls back to DES

I suppose that's why libratbox has its own crypt() implementation, because that's a mighty good way they decided to deprecate MD5, with a noticeable improvement in security. Arsehats.

(I don't really like crypt myself, but I also don't like removing the upgrade path for old accounts.)

dwfreed:
I mean Atheme falls back to DES mode.  This likewise wouldn't work in OpenBSD.

ilbelkyr:
@aaronmdjones you've been working on this – opinions on falling back to the `!HAVE_CRYPT` implementation for OpenBSD? We should encourage pbkdf2v2 or argon2d for new installations, but someone with existing `crypto/posix` registrations might want to move their installation to OpenBSD.

(Also, as per @grawity, `crypt` per POSIX *can* return `NULL` in case of error, specifically an `errno` value of `ENOSYS` aka "function not implemented", so we probably could do with a check for that rather than just checking for OpenBSD.)

aaronmdjones:
@ilbelkyr given that `libathemecore/md5` is always available regardless of platform, I could just switch to using that unconditionally if the digest looks like an MD5 one.

Also on my TODO for this module though is to add SHA-2 support (e.g. GNU libc `$5$` and `$6$`), ideally before we release 7.3.0. This would allow people to keep using the `posix` module if they want to, but digests would be upgraded as people identify or set new passwords, similar to raising the value of `pbkdf2v2::digest`. It won't be nearly as strong (resistant to cracking if the database is compromised) as the `pbkdf2v2` or `argon2d` modules, but it will be better.

One thing that comes to mind there, though, is what to name it? It obviously won't be `posix` any more, since `crypt(3)` has no mention of SHA-2 (or even MD5)...

aaronmdjones:
Closing as we have no intention of merging this; we will fix it instead. I have already fixed the segfault issue (commit 317796ead0ea5f34d4a9) and the POSIX module issue will be addressed in issue #603

PR #522, opened 2016-10-19
Release/7.2

Indjov:
I make a Bulgarian translation of services.

email : al@indjov.info
uploaded file : http://www.filedropper.com/bg_5

10x in advance


grawity:
that's not how pull requests work, but ok

PR #538, opened 2017-02-03
Refactor how sourceinfos are rendered

mniip:
Add a method to the sourceinfo vtable to let it override its rendition


mniip:
Will have to be rebased if #539 is to be merged.

PR #547, opened 2017-03-19
Added cyrillic support

Arkadietz:
Added cyrillic support

### DIFF
--- a/include/match.h
+++ b/include/match.h
@@ -27,6 +27,8 @@ E int match_mapping;
 #define C_DIGIT 0x00000002
 #define C_NICK  0x00000004
 #define C_USER  0x00000008
+#define CYR_C   0x40000 /* Added cyrillic support by Rix @ 27 May 2016 */
+#define NALPHA_C 0x80000 /* Added cyrillic support by Deyan Hristanov */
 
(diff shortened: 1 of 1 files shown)


grawity:
Sorry, this kind of "support" is not suitable for upstream.

The main problem is that your patch essentially changes Atheme behavior for *everyone*, even if they would prefer an entirely different codepage. It wasn't made configurable; it doesn't have an offer; and you haven't even explained *which* codepage it is for, although I'm guessing Windows-1251.

While I would love for IRC to have better internationalization support, it needs to be done properly, not by favoring an ad-hoc hack for a single language. If we accepted this kind of patch for cp1251, what would we say to cp1252 or cp1257 or iso2022 users?

At minimum, if support for single-byte charsets is needed, then you would have to make it a configurable option so that multiple character tables could be added. (For example, inspircd's nationalchars module...)

But honestly, it would be better to spend time on Unicode/UTF-8 support instead. IRCv3 already has a discussion about that.

---

As for your Git usage: While it's good to split up a feature across several commits, keep all of them in a single pull request. Conversely, while Atheme _does_ need indentation cleanup, that should be left to an entirely separate PR.

PR #552, opened 2017-04-12
gameserv: Use arc4random_uniform() everywhere

grawity:
It compiles.


siniStar7boy:
Thanks @grawity - will boot up the "test machine" in a bit and run through some testing and then merge if all goes well.

PR #562, opened 2017-06-16
MAGIRC Network stats

Indjov:
Hello, 
What do you think about putting this statistic into the atema. Denora since it is no longer supported, it is now integrated into Anope Services , and they use MagIRC like webinterface.


grawity:
If you want to open an issue, then open an issue, not a pull request with half the repository in it...

Indjov:
@grawity , I`m so sorry but i try to open issue from my phone. I don`t have any idea how i push the pull request.

I'm sorry for my ignorant experience. 
I apologize again.

PR #565, opened 2017-06-27
Notify users if verification is incomplete

Gryllida:
Notify users who just identified, if verification is incomplete. Untested edit. Copied from modules/nickserv/info.c.

### DIFF
--- a/modules/nickserv/identify.c
+++ b/modules/nickserv/identify.c
@@ -169,7 +169,10 @@ static void ns_cmd_login(sourceinfo_t *si, int parc, char *parv[])
 		}
 
 		command_success_nodata(si, nicksvs.no_nick_ownership ? _("You are now logged in as \2%s\2.") : _("You are now identified for \2%s\2."), entity(mu)->name);
-
+		
+		if (MU_WAITAUTH & mu->flags)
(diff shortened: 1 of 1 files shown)


modules/nickserv/identify.c
```diff
@@ -169,7 +169,10 @@ static void ns_cmd_login(sourceinfo_t *si, int parc, char *parv[])
 		}
 
 		command_success_nodata(si, nicksvs.no_nick_ownership ? _("You are now logged in as \2%s\2.") : _("You are now identified for \2%s\2."), entity(mu)->name);
-
+		
+		if (MU_WAITAUTH & mu->flags)
+			command_success_nodata(si, _("You have \2NOT COMPLETED\2 registration verification"), entity(mu)->name);
```
GLolol: Nitpick: missing a period at the end? (the other messages are all sentences)

ilbelkyr:
Assuming we're going to do this, it might be a good idea to do this for any sort of identification, not just `nickserv/identify`. (This PR won't catch e.g. SASL logins.)

I think we have a `user_identify` hook that you could hook into (probably from the module doing the verification).

PR #566, opened 2017-06-28
Notify users if verification is incomplete

Gryllida:
via a hook in verify

### DIFF
--- a/modules/nickserv/verify.c
+++ b/modules/nickserv/verify.c
@@ -15,6 +15,7 @@ DECLARE_MODULE_V1
 	VENDOR_STRING
 );
 
+static void on_user_identify(user_t *u);
 static void ns_cmd_verify(sourceinfo_t *si, int parc, char *parv[]);
 static void ns_cmd_fverify(sourceinfo_t *si, int parc, char *parv[]);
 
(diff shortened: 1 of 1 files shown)


modules/nickserv/verify.c
```diff
@@ -235,6 +238,15 @@ static void ns_cmd_fverify(sourceinfo_t *si, int parc, char *parv[])
 	}
 }
 
+static void on_user_identify(user_t *u)
+{
+	myuser_t *mu = u->myuser;
+
+	if (MU_WAITAUTH & mu->flags)
+		notice(nicksvs.nick, u->nick, _("You have \2NOT COMPLETED\2 registration verification."));
+		//command_success_nodata(si, _("You have \2NOT COMPLETED\2 registration verification."), entity(mu)->name);
```
Zarthus: This comment is unnecessary, especially when you consider the way the if-statement is written together with the indentation, someone uncommenting this will trigger this hook every time.

ilbelkyr:
FWIW, you should delete the hook in `_moddeinit`. That said, I think it's easier (and perhaps a little cleaner) to add the check to `myuser_login` directly.

PR #576, opened 2017-08-19
SASL: Use a parameter vector to allow arbitrary number of S2S arguments

aaronmdjones:
This can be used in future to restrict the authentication mechanisms
that plaintext clients are allowed to perform (for example, to
prevent password disclosure to passive observers by preventing
SASL PLAIN).

c.f. https://github.com/charybdis-ircd/charybdis/commit/754c1edf2e9c

### DIFF
--- a/modules/protocol/inspircd.c
+++ b/modules/protocol/inspircd.c
@@ -1347,11 +1347,18 @@ static void m_encap(sourceinfo_t *si, int parc, char *parv[])
 		if (parc < 6)
 			return;
 
+		(void) memset(&smsg, 0x00, sizeof smsg);
+
 		smsg.uid = parv[2];
 		smsg.mode = *parv[4];
-		smsg.buf = parv[5];
-		smsg.ext = parc >= 6 ? parv[6] : NULL;
-		smsg.server = si->s ? si->s : NULL;
+		smsg.parc = parc - 5;
+		smsg.server = si->s;
+
+		if (smsg.parc > SASL_MESSAGE_MAXPARA)
+			smsg.parc = SASL_MESSAGE_MAXPARA;
+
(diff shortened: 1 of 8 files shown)


aaronmdjones:
CC @MrBenC

aaronmdjones:
(In case anyone is wondering, I have confirmed that SASL login still works with these changes)

```
-> :42X ENCAP services.int SASL 42XAAAAAC 00AAAAAAE H 127.0.0.1 127.0.0.1 S
-> :42X ENCAP services.int SASL 42XAAAAAC 00AAAAAAE S PLAIN
<- :00A ENCAP saiga.int SASL 00AAAAAAE 42XAAAAAC C +
-> :42X ENCAP services.int SASL 42XAAAAAC 00AAAAAAE C QWFyb24AQWFyb24AcGFzc3dvcmQ=
<- :00A ENCAP saiga.int SVSLOGIN 42XAAAAAC * * * Aaron
<- :00A ENCAP saiga.int SASL 00AAAAAAE 42XAAAAAC D S
-> :42X EUID aaron 1 1503125083 +Zi ~aaron 127.0.0.1 127.0.0.1 42XAAAAAC * Aaron :Aaron Jones
```

modules/protocol/inspircd.c
```diff
@@ -1347,11 +1347,18 @@ static void m_encap(sourceinfo_t *si, int parc, char *parv[])
 		if (parc < 6)
 			return;
 
+		(void) memset(&smsg, 0x00, sizeof smsg);
```
ilbelkyr: A quick search suggests we tend to avoid casting the `memset()` return to `void`. As for `0x00`, it looks like most existing code varies between `0` and `'\0'` (though pbkdf2v2 has `0x00` as well – perhaps unsurprisingly, given that as I recall, you wrote that module); I don't really mind, though.

PR #584, opened 2017-10-09
Advertise SASL mechanisms to UnrealIRCd servers.

syzop:
We added support for SASL v3.2 in UnrealIRCd which requires services to send the saslmechlist. I took a look at how you do it with other IRCd's and made a patch.


grawity:
Compared to other sts messages it seems to be missing the `:servername` prefix, is that intended?

syzop:
You are free to send it but it wouldn't add any value.

syzop:
Thanks :)

PR #585, opened 2017-10-11
fix some syntax errors

Indjov:
fix some syntax errors

### DIFF
--- a/po/bg.po
+++ b/po/bg.po
@@ -1,270 +1,83 @@
-# Copyright (C) 2004 - 2011 Atheme
-# This file is distributed under the same license as the PACKAGE package.
-# Translators:
-# Ignacio Riquelme Morelle <shadowm2006@gmail.com>, 2010, 2011.
-# sanad <0sanad0@gmail.com>, 2013.
 msgid ""
 msgstr ""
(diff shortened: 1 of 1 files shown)


Zarthus:
Your diff is extremely large in in some situations removes translations. I don't think this is intended.

po/bg.po
```diff
@@ -1,270 +1,83 @@
...
-msgstr "%s е замразен."
+msgstr ""
 
-#: modules/botserv/bottalk.c:86 modules/botserv/bottalk.c:152
+#: modules/botserv/bottalk.c:86 modules/botserv/bottalk.c:152 
 #: modules/botserv/main.c:954
 #, c-format
 msgid "%s does not have a bot assigned."
-msgstr "%s няма назначен бот."
+msgstr ""
```
Zarthus: Translation was removed. (I see at least 9 different cases in here without inflicting self harm due to browser speed)

ilbelkyr:
@Indjov any chance you could address @Zarthus's comment? I'm not familiar with the language myself so can't really say much more than he already did.

Indjov:
@ilbelkyr , i will make a new translate.

PR #586, opened 2017-10-18
ChatServices Service Pack I

MrBenC:
This pull request contains a number of changes that I consider important and would really like to see be included in 7.2.10 or 7.3.0, whichever version the next stable release will have.

At an absolute minimum:
- The PUBACL bugs, which currently render it defacto useless since it won't be saved across services restarts.
- ~~The send_sasl_quit stuff, which resolves a problem on some IRCds where if cap-notify is enabled, the message about the loss of the sasl cap is not passed on to clients that support cap-notify.~~  Upon further reflection, some comments, and some further discussion, this is a bit of a hack.  It doesn't fully solve the problem and it really should be fixed in the affected IRCds.  I'll make a pull request for the affected IRCds (that I'm aware of) very soon.

### DIFF
--- a/dist/atheme.conf.example
+++ b/dist/atheme.conf.example
@@ -2206,6 +2206,14 @@ general {
 	 */
 	immune_level = immune;
 
+	/* (*)send_sasl_quit
+	 * If enabled, SaslServ will send a /QUIT before the the services server
+	 * either restarts or shuts down.  Some IRCds unfortunately need this in
+	 * order for cap-notify and the "sasl" cap to function properly.
+	 * Default: disabled
+	 */
+	#send_sasl_quit;
+
 	/* show_entity_id
 	 * This makes nick/user & group entity IDs visible to everyone, rather
 	 * than just opers with user:auspex or group:auspex privileges.
--- a/dist/atheme.conf.example
+++ b/dist/atheme.conf.example
@@ -2206,14 +2206,6 @@ general {
 	 */
 	immune_level = immune;
 
-	/* (*)send_sasl_quit
-	 * If enabled, SaslServ will send a /QUIT before the the services server
-	 * either restarts or shuts down.  Some IRCds unfortunately need this in
-	 * order for cap-notify and the "sasl" cap to function properly.
-	 * Default: disabled
-	 */
-	#send_sasl_quit;
-
 	/* show_entity_id
 	 * This makes nick/user & group entity IDs visible to everyone, rather
 	 * than just opers with user:auspex or group:auspex privileges.
--- a/dist/atheme.conf.example
+++ b/dist/atheme.conf.example
@@ -50,6 +50,7 @@
  * Asuka 1.2.1 or later                         modules/protocol/asuka
  * Bahamut 1.8.x                                modules/protocol/bahamut
  * Charybdis IRCd                               modules/protocol/charybdis
+ * ChatIRCd                                     modules/protocol/chatircd1.1
  * DreamForge 4.6.7 or later                    modules/protocol/dreamforge
  * InspIRCd 2.0                                 modules/protocol/inspircd
  * ircd-ratbox 2.0 and later                    modules/protocol/ratbox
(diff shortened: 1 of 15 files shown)


dist/atheme.conf.example
```diff
@@ -2206,6 +2206,14 @@ general {
 	 */
 	immune_level = immune;
 
+	/* (*)send_sasl_quit
+	 * If enabled, SaslServ will send a /QUIT before the the services server
+	 * either restarts or shuts down.  Some IRCds unfortunately need this in
+	 * order for cap-notify and the "sasl" cap to function properly.
```
grawity: Which ircds are those? Doesn't the services server disconnecting already imply a /quit for all its pseudo-clients?

janicez:
Just looking at the commit messages, this looks like a good 7.3 SP 0 or 7.2 SP 11 patchset.

It seems to fix oversights without being clearly additive of functionality that wasn't meant to be there. The only thing I do not condone is 'sasl-quit', which is quite bizarre (should not services be sending these anyway if it's a known problem?) and including the Chatircd module.

MrBenC:
@janicez No.. I believe it's a reasonable assumption for the IRCd to count splits as quitting.  Services shouldn't actually need to send a special /quit message prior to shutdown or restart.  If it doesn't, perhaps it may be a bug in the IRCd.

I'm including the option because I don't know if there may or will be other IRCds that suffer from this problem.

GLolol:
I think it's more straightforward and future-proof to just fix this in the affected IRCds.

dist/atheme.conf.example
```diff
@@ -50,6 +50,7 @@
  * Asuka 1.2.1 or later                         modules/protocol/asuka
  * Bahamut 1.8.x                                modules/protocol/bahamut
  * Charybdis IRCd                               modules/protocol/charybdis
+ * ChatIRCd                                     modules/protocol/chatircd1.1
```
maxteufel: Who uses this? Is it really necessary in Atheme? A separately maintained version also has its benefits: independence from Atheme development and we're not the ones who need to fix the issues with it.

siniStar7boy: I know of at least a handful of networks, including mine, that run ChatIRCD.

modules/protocol/chatircd1.1.c
```diff
@@ -0,0 +1,352 @@
+/*
```
maxteufel: There seems quite a lot of code duplication from protocol/charybdis.c here. That can be changed similar to the ircd-seven module (rather than ts6-generic, depend on charybdis in `_modinit`, then update what only what needs to be changed).

maxteufel:
@MrBenC I think this would be easier to handle if it wasn't a huge pull request, fwiw.

modules/protocol/chatircd1.1.c
```diff
@@ -0,0 +1,352 @@
...
+#include "atheme.h"
+#include "uplink.h"
+#include "pmodule.h"
+#include "protocol/chatircd.h"
+
+DECLARE_MODULE_V1
+(
+	"protocol/chatircd", true, _modinit, NULL,
+	PACKAGE_STRING,
+	"ChatLounge IRC Network Development Team <http://www.chatlounge.net>"
```
Zarthus: https perhaps?

maxteufel: Nope, this should be `VENDOR_STRING`, as in all other modules. Since it is in fact, a vendor, not a developer string, Atheme is the appropriate content for this.

MrBenC:
I have updated the protocol module to reduce code duplication to the extent possible and used VENDOR_STRING in the new module.  This should hopefully resolve any outstanding problems.

modules/protocol/chatircd1.1.c
MrBenC: No longer a relevant concern, as the protocol module has been rewritten to resolve this.

PR #587, opened 2017-10-18
modules/: update to VENDOR_STRING macro where not done yet

maxteufel:


ilbelkyr:
This is now covered by #595 which has been merged.

PR #593, opened 2017-10-31
Add a couple login-related hooks

mniip:
* Add a user_can_logout(si, u) hook that can prevent a user from logging out

  In a few dubious cases where it would be incorrect/impossible to deny the operation, the target user is killed instead (as if on P10).

* Add a user_can_rename(si, mu, mn) hook

### DIFF
--- a/modules/nickserv/identify.c
+++ b/modules/nickserv/identify.c
@@ -61,8 +61,6 @@ static void ns_cmd_login(sourceinfo_t *si, int parc, char *parv[])
 	const char *target = parv[0];
 	const char *password = parv[1];
 	char lau[BUFSIZE];
-	hook_user_login_check_t login_req;
-	hook_user_logout_check_t logout_req;
 
 	if (si->su == NULL)
 	{
@@ -92,12 +90,17 @@ static void ns_cmd_login(sourceinfo_t *si, int parc, char *parv[])
 		return;
 	}
 
-	login_req.si = si;
-	login_req.mu = mu;
-	login_req.allowed = true;
-	logout_req.si = si;
-	logout_req.u = u;
-	logout_req.allowed = true;
+	hook_user_login_check_t login_req = {
+		.si      = si,
+		.mu      = mu,
+		.allowed = true,
+	};
+	hook_user_logout_check_t logout_req = {
+		.si      = si,
+		.u       = u,
+		.allowed = true,
+		.relogin = true,
+	};
 	hook_call_user_can_login(&login_req);
 	if (login_req.allowed && u->myuser)
 		hook_call_user_can_logout(&logout_req);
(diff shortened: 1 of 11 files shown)


modules/nickserv/identify.c
```diff
@@ -90,11 +92,17 @@ static void ns_cmd_login(sourceinfo_t *si, int parc, char *parv[])
...
+	login_req.mu = mu;
+	login_req.allowed = true;
+	logout_req.si = si;
+	logout_req.u = u;
+	logout_req.allowed = true;
+	hook_call_user_can_login(&login_req);
+	if (login_req.allowed && u->myuser)
+		hook_call_user_can_logout(&logout_req);
+
+	if (!login_req.allowed || !logout_req.allowed)
```
Zarthus: Wouldn't the text beneath be wrong here? If I am not permitted to log out services would send me "You cannt log in as %s because ..."

maxteufel: This is about logging in via `/ns identify` here. If you are already identified, it checks whether you are permitted from logging out of the old account and then logging in again. So technically the text is correct.

The actual question would be though, if this should be classified as "logging out, then logging in" or actually "change account (name)".

modules/nickserv/logout.c
```diff
@@ -58,7 +61,27 @@ static void ns_cmd_logout(sourceinfo_t *si, int parc, char *parv[])
...
+	}
 
+	req.si = si;
+	req.u = u;
+	req.allowed = true;
+	hook_call_user_can_logout(&req);
+	if (!req.allowed)
+	{
+		logcommand(si, CMDLOG_LOGIN, "failed LOGOUT \2%s\2 (denied by hook)", u->nick);
+		command_fail(si, fault_authfail, _("You cannot logout \2%s\2 because the server configuration disallows it."), entity(u->myuser)->name);
```
maxteufel: If you want to keep this consistent with the text in nickserv/identify it should be `log out`, not `logout` (since there it is `log in`).

modules/nickserv/identify.c
ilbelkyr: I wonder if it's worth providing the `user_can_logout` hook with an additional parameter specifying whether the "logout" will actually leave the user not logged in to any account (as opposed to merely changing accounts).

mniip:
That is a good idea, but not quite sure what to do about it. In principle a module could hook `user_can_login` and depend on the fact that `user_can_logout` is invoked immediately after it with the same sourceinfo.

mniip:
Another aspect is that the best we can specify is that the user is *trying to* login - this hook isn't the last check (for good reasons) and the user can still fail to login after it.

ilbelkyr:
Right, but if the user fails to log in, they will be left on their old account, hence we can ask "may this user log out to identify to another account" safely, I think.

(Or to amend my previous comment: I wonder if it's worth providing the `user_can_logout` hook with an additional parameter specifying whether the attempted action **(if successful)** will actually leave the user not logged in to any account (as opposed to merely changing accounts). **If the action is unsuccessful, their login state will remain the same anyway.**)

PR #595, opened 2017-11-05
Module fixups

aaronmdjones:
```
modules: fix up declarations

While investigating something unrelated the other day, I stumbled
upon this quirk in module declarations.

The second field of DECLARE_MODULE_V1() was supposed to mean
"is a core module?" (can't be reloaded), or so I thought.

It turns out that this field is used to initialise the 6th field
of the MAPI module header symbol, and that field is an
'module_unload_capability_t' enumeration, not a boolean.

The compiler doesn't warn for this because bool in C is a kludge,
so when you're passing 'false' you're really passing '0', which
matches MODULE_UNLOAD_CAPABILITY_OK, and when you're passing
'true' you're really passing '1', which matches
MODULE_UNLOAD_CAPABILITY_NEVER.

Still, it is a discontinuity that should be addressed before it
confuses someone else for a moment too.





modules: fix up formatting

- Rewrap DECLARE_MODULE_V1 lines to this format:

  DECLARE_MODULE_V1("name", module_unload_capability_t,
                    _modinit, _moddeinit,
                    "module_version", "module_vendor");

  (the reason why will become clear in a subsequent commit)

- Remove trailing vim options detritus
- Trim file headers
- Remove some lines consisting entirely of whitespace





modules: declaration macro improvements

- Add a VENDOR_DECLARE_MODULE_V1 macro which uses PACKAGE_STRING
  as the module version

- Add a SIMPLE_DECLARE_MODULE_V1 macro which uses VENDOR_STRING
  as the module vendor, and PACKAGE_STRING as the version

  (Most modules have these 2 arguments)

- Convert all in-tree modules to use the 2 new macros

  The old DECLARE_MODULE_V1 macro will be kept around for
  compatibility with contrib modules and out-of-tree modules

Also remove yet more $EDITOR detritus missed in the last commit





modules: initialisation/deinitialisation cleanup

The initialisation and deinitialisation routines should be static
as they are only expected to be accessed from within the MAPI header
created by the DECLARE_MODULE_V1() family of macros.





modules: module declaration cleanups

For modules that declare a vendor of us, just use the new simple
declaration macro instead (which does the same thing, but with
a single shared value, so it doesn't get out of sync with other
modules if e.g. our URI changes)
```


aaronmdjones:
Note that this pull request conflicts with commit 3db375030c7df6c1948b (PR #587) but I can refactor everything to use `VENDOR_STRING` if you'd like.

aaronmdjones:
Bump.

ilbelkyr:
Looks good to me; this does end up moving the module declaration to the bottom of the file but I don't really object to that.

I'm not sure what we should do with the "vendor" field. Some modules use it more like an "author" field. I suspect modules in the atheme/atheme repo should all default to the same vendor string. Thoughts?

aaronmdjones:
I don't have any opinion one way or the other, but I can make everything in-tree claim a vendor of us quite easily if that makes the most sense to you.

ilbelkyr:
Yeah, given the reactions on #587 that seems like a good course of action.

aaronmdjones:
I do believe this is ready for merging.

PR #598, opened 2017-12-13
SASL Code Quality and Feature Improvements

aaronmdjones:
The `saslserv/main` commits that end in 'move' only move functions around; they don't change any code. This should speed up review.


modules/saslserv/authcookie.c
```diff
@@ -11,44 +11,45 @@
...
+	const char *const authzid = ptr;
+	if (! *ptr || (ptr += strlen(ptr) + 1) >= end)
 		return ASASL_FAIL;
-	strcpy(authz, message);
-	message += strlen(message) + 1;
 
-	/* Copy the authcid */
-	if(strlen(message) > 255)
+	const char *const authcid = ptr;
+	if (! *ptr || (ptr += strlen(ptr) + 1) >= end)
```
grawity: Really not sure whether I like assignments hidden so deep inside an `if()`.

aaronmdjones: The alternative is a lot more lines that do the same thing; a la

```
    if (! *ptr)
        return ASASL_FAIL;

    ptr += strlen(ptr) + 1;
    if (ptr >= end)
        return ASASL_FAIL;
```

That would get old, quickly, no?

grawity: Eh, I think that'd be fine. Yes, I guess you can't fit as much in a single screen, but (as someone who isn't all that fluent in C) it's easier to decipher at a quick glance – specifically, it's easier to notice that `ptr` is being assigned to.

`||` chains are fine, assignments inside if() are okay I suppose, but _combining_ the two begins to look like golfing just for the sake of it... How about this then:

``` c
const char *const authzid = ptr;

if (! *ptr)
    return ASASL_FAIL;
if ((ptr += strlen(ptr) + 1) >= end)
    return ASASL_FAIL;

const char *const authcid = ptr;
```

(Maybe that first check should in fact be `if (! *authzid)`? Technically it's the same thing, but semantically it makes more sense.)

aaronmdjones: How about

```C
const char *const authzid = ptr;

if (! *authzid)
    return ASASL_FAIL;
if ((ptr += strlen(authzid) + 1) >= end)
    return ASASL_FAIL;

const char *const authcid = ptr;

// ...
```

grawity: LGTM

aaronmdjones: https://github.com/atheme/atheme/pull/598/commits/8c64dce849a0ec6003e8f9b0874d4ef395a8bb3c

aaronmdjones:
I do believe I am done... unless any concerns are raised here.

aaronmdjones:
I have discovered a bug in `sasl_input()` that has been there forever (as far as I can tell) and was not accidentally corrected by my refactoring. The following commit should resolve the issue and then the commit after that removes some (now) duplicate code.

aaronmdjones:
Apologies for GitHub; it put the 2 aforementioned commits *BEFORE* my comment.

modules/saslserv/main.c
```diff
@@ -656,11 +622,11 @@ sasl_newuser(hook_user_nick_t *data)
 	p->flags &= ~ASASL_NEED_LOG;
 
 	/* Find the account */
-	myuser_t *const mu = p->authzid ? myuser_find_by_nick(p->authzid) : NULL;
+	myuser_t *const mu = p->authzeid ? myuser_find_uid(p->authzeid) : NULL;
 	if (! mu)
 	{
 		(void) notice(saslsvs->nick, u->nick, "Account %s dropped, login cancelled",
-		                                      p->authzid ? p->authzid : "??");
+		                                      p->authzeid ? p->authzeid : "??");
```
ilbelkyr: As remarked on IRC, this should possibly indicate the client-specified authzid instead. We would probably need to keep track of it alongside the EID for logging purposes.

modules/saslserv/main.c
```diff
@@ -739,6 +705,64 @@ sasl_mech_unregister(struct sasl_mechanism *mech)
...
+	(void) hook_call_user_can_login(&req);
+
+	if (! req.allowed)
+		(void) logcommand(p->si, CMDLOG_LOGIN, "failed LOGIN to \2%s\2 (denied by hook)", entity(mu)->name);
+
+	return req.allowed;
+}
+
+static bool
+sasl_authzid_can_login(struct sasl_session *const restrict p, const char *const restrict authzid,
```
ilbelkyr: I'll admit the code duplication between `sasl_authzid_can_login` and `sasl_authcid_can_login` makes me a little unhappy.

ilbelkyr:
(Not really related to this PR, but something I thought of while reviewing it – doesn't `saslserv/scram-sha` need to be taught about the `MU_NOPASSWORD` flag?)

Issue #608, opened 2018-01-12
MemoServ read back sent memos


Zarthus:
It's easy to miss memos, and even easier to forget what you sent to someone.

A way to read back memos you sent for the eventual case where someone responds to your several-{day,week,month,year}s old memo would be really nice.

PR #609, opened 2018-01-12
memoserv/read: display text of a read memo

siniStar7boy:
Display the original text in a memo to the user when notifying that their (sent) memo has been read. (Reference atheme/atheme#608)


maxteufel:
This pull request should be against master, not release/7.2. Anyway, see the improved new pull request.

PR #610, opened 2018-01-12
memoserv/read: send back original text with read notification

maxteufel:
It is easy to forget what memos you sent months, even years, ago that
some user read just now. Hence, we now send back the original memo with
the read receipt notification.

Original code taken from atheme/atheme#609 and Xtheme changes.

Closes atheme/atheme#608 ("MemoServ read back sent memos")

### DIFF
--- a/modules/memoserv/read.c
+++ b/modules/memoserv/read.c
@@ -30,8 +30,8 @@ static void ms_cmd_read(sourceinfo_t *si, int parc, char *parv[])
 {
 	/* Misc structs etc */
 	myuser_t *tmu;
-	mymemo_t *memo, *receipt;
-	mowgli_node_t *n;
+	mymemo_t *memo, *receipt, *original;
+	mowgli_node_t *n, *m;
 	unsigned int i = 1, memonum = 0, numread = 0;
 	char strfbuf[BUFSIZE];
 	struct tm tm;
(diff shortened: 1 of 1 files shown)


modules/memoserv/read.c
```diff
@@ -92,11 +92,14 @@ static void ms_cmd_read(sourceinfo_t *si, int parc, char *parv[])
 				if (memo->status & MEMO_CHANNEL)
 					;
 				else if (strcasecmp(si->service->nick, memo->sender) && (tmu != NULL) && (tmu->logins.count > 0))
```
Zarthus: Just to recap what we talked about IRC (and add my own thoughts);

Due to the way the comparison currently works it's possible for someone to be using your nick but not your account to obtain your memo.

This exploit kind of relies on many factors though, primarily that you're actually using MemoServ to send sensitive memos. (which is quite hard to do with a 400 character limit..)

The solution could be to store the EUID with memos so you can do comparisons against the account identifier itself rather than the string.

```
(19:14:50) <@mt> 18:39:35 <+Zarthus> theoretical thought: is it possible to listen in on someone elses memo by using their name that way?
(19:15:18) <@mt> yeah, it seems that is possible for non-authenticated senders or accountname changes/dropped nicks
(19:15:34) <+amdj> is the EID stored? could use that
(19:18:54) <@mt> amdj: memo->sender is stored as string of an entity name. EID is not stored
(19:20:12) <+amdj> might be worth storing the sender's EID in 7.3.0 instead. this would necessitate a DBV bump and some small migration logic for 7.2 databases
(19:21:59) <+amdj> only problem is what if they drop their account after sending the memo but before it's read ... could put [nick] at the start of the memo text or store the EID as well as their nick
(19:32:50) <@mt> what if we just leak the EID in the case that it can't be resolved to an account anymore, that shouldn't be too much of a concern anyway
(19:49:25) <@siniStar> +1
```

PR #611, opened 2018-01-13
docs: clarify Git-related messages

grawity:


GIT-Access.txt
```diff
@@ -0,0 +1,31 @@
+Building from the Git repository
+--------------------------------
+
+The Atheme Git repository can be checked out using the following command:
+
+	$ git clone https://github.com/atheme/atheme.git atheme-devel
```
Zarthus: Is there a reason not to recommend `git clone [...] --recursive` outright?

grawity: Probably not; is it available in all git versions?

Zarthus: Since [1.6.5](https://github.com/git/git/blob/master/Documentation/RelNotes/1.6.5.txt#L102)

Debian 7 (Wheezy, oldoldstable) ships 1.7.x

Unless they're running something like RHEL, yes.

aaronmdjones:
Closing as I merged a slightly modified version just now.

Issue #603, opened 2017-12-27
modules/crypto/posix needs some serious love


aaronmdjones:
There are a few concerns with this module:

1) The `posix_md5crypt` function needs general cleanup, more comments for its operations, style fixes (inconsistent indentation), ...

2) We should investigate using `libathemecore/md5` for all digests that look like MD5 digests (`$1$...`) on some platforms that do not implement it properly in `crypt(3)` (e.g. OpenBSD cf. issue #518) even if `HAVE_CRYPT` is true. This can be tested for in `mod_init` with some reference test vectors.

3) We should strongly consider implementing support for better digest algorithms (e.g. SHA2-256 a la `$5$...` and SHA2-512 a la `$6$...`) so that we can both:

   1) Verify more password hashes from other implementations or data migration sources
   2) Upgrade password digests over time as people identify or set new passwords

**NOTE**: If item 3 ever bears fruit we should also consider renaming the module because there is no mention of SHA2 in any POSIX `crypt(3)` manual page; an alternative would be to create `modules/crypto/sha256crypt` & `modules/crypto/sha512crypt` or the like. Such modules could also not strictly *need* OpenSSL (libcrypto) if we add a SHA-2 implementation to libathemecore; OpenSSL would only provide a (negligible, in this context) speed boost and a more tested interface, but the latter can be easily verified with reference test vectors and regular interoperability tests.

A nice side-benefit of adding SHA to libathemecore would be that `modules/crypto/pbkdf2` and `modules/crypto/pbkdf2v2` would also no longer need OpenSSL, because once you have the digest algorithms implemented, adding HMAC and PBKDF2 on top is simple and easy.

PR #614, opened 2018-01-20
Implement a unified message digest API

aaronmdjones:


aaronmdjones:
The only part of this that has not been tested so far is SASL SCRAM-SHA logins with the ARM mbedTLS frontend; everything else is feature-complete and works -- I will be testing mbedTLS on my own testnet later today.

aaronmdjones:
Can anyone think of any other common cryptographic libraries that I can add support for?

The following are required:

1. Supports MD5, SHA1, SHA2-256 and SHA2-512
2. Is not an experience in hell to write code for

The following are not required, but are considered bonuses:

3. Avoids dynamic memory allocation
4. Supports HMAC
5. Supports PBKDF2-HMAC

The only one that comes to mind is libgcrypt, but that may violate required item 2. At least, when I glanced at the man pages when I was writing this, I didn't like what I saw.

grawity:
[nettle](https://www.lysator.liu.se/~nisse/nettle/nettle.html)? (It's the underlying crypto library of GnuTLS so relatively common.) No personal experience with it though.

aaronmdjones:
How on earth did I forget about nettle ...

I'll have a look at that tomorrow. Thanks.

aaronmdjones:
Apologies, my trackpad hit the wrong button.

aaronmdjones:
Another issue is that at the moment the configure script will decide to use ARM mbedTLS as the digest frontend if the libmbedcrypto library is found and is usable, but the necessary headers may not be available.

I will extend the detection to include headers tomorrow too, like it currently does for OpenSSL.

aaronmdjones:
All of the digest frontends pass our integrated test suite.

aaronmdjones:
I do believe I am done unless anything is raised here.

aaronmdjones:
Note that `modules/crypto/posix` has deliberately not been modified because I'd rather rewrite it (#603).

PR #621, opened 2018-03-02
Atheme spring cleaning

aaronmdjones:


aaronmdjones:
Note that I still need to put this on my testnet. I will do that this weekend.

aaronmdjones:
GitHub doesn't seem to be rendering the timeline or complete list of commits accurately, so you'll want to

```
$ git log master..atheme-spring-cleaning
```

aaronmdjones:
Also note that contrib doesn't work at the moment. I'm working on that right now.

aaronmdjones:
Contrib should work now (it compiles, at least). Most of the changes are very minimal, but I do need to test `os_resolve.c` and `dnsbl.c`. I will do so on my testnet this weekend.

aaronmdjones:
The modified contrib modules work. I even found a bug in libmowgli because of it (atheme/libmowgli-2@21658b28f1323f11f1aa).

aaronmdjones:
GitHub is hiding most of the commits and all of my previous comments by default, which is highly annoying.

To reiterate one of them, it's not even displaying (hidden or not) most of the commits, and the ones it is, is showing them out of order. To see what changed in commit order, you'll want to

` $ git log master..atheme-spring-cleaning`

PR #627, opened 2018-06-06
Licensing & copyright updates

aaronmdjones:
Entire Project: Clean up licensing, copyrights, etc.

For files licensed under `ISC` (those that refer to `doc/LICENSE`, which is `ISC`, and/or include the `ISC` license permissions text), replace the reference to `doc/LICENSE` (if any) with an `SPDX-License-Identifier: ISC` tag, and remove the disclaimer (as there are several variations of it, while only the text on the license webpage and in `doc/LICENSE` is sufficient, and the permissions do not prohibit its removal, unlike with e.g. the `BSD 3-Clause` license).

For files licensed under `BSD 2-Clause` and `BSD 3-Clause`, the disclaimer cannot be removed, as it must be retained to comply with the license. Still, add an `SPDX-License-Identifier` tag for the relevant license.

For files located in the public domain, add an `SPDX-License-Identifier: CC0-1.0` tag.

For every file that gets a tag, also include a link to the relevant license(s).

Some of these commits will also add missing copyright headers to files that I have modified over the last year or so, and some of them will relicense some of my existing contributions from `BSD 3-Clause` to `ISC`.

Finally, some of these commits will rename a URI-less `Atheme Development Group` to `Atheme Project (http://atheme.org/)`, which was the name used for the developers back then, consistent with the convention used under `libathemecore/*.c` since forever.

To more clearly distinguish the previous stewards of this codebase from the current ones, the current `Atheme Development Group` will have the text `(https://atheme.github.io/)` appended to the end if it was missing, or the URI updated if it was not.

Where files have been modified since the project was handed over, and the copyright year (mistakenly) bumped on the existing author, split it into 2 separate attributions instead, as not all members of one
group are also members of the other.


libathemecore/version.c
```diff
@@ -1,82 +1,98 @@
...
-  "  nenolod, William Pitcock <nenolod@nenolod.net>",
-  "  rakaur, Eric Will <rakaur@malkier.net>",
-  "  rintaun, Matt Lanigan <rintaun@projectxero.net>",
-  "  sycobuny, Stephen Belcher <sycobuny@malkier.net>",
-  " ",
-  "Visit our website at http://atheme.github.io",
-  0,
+	" ",
+	"Atheme IRC Services --",
+
```
ilbelkyr: Hm, there's a blank line here that won't be in the output; everywhere else in the text, blank lines in the source only appear where there will be blank lines in the output as well. Just a minor nitpick, though.

libathemecore/version.c
```diff
@@ -1,82 +1,98 @@
...
+	"  kaniini, William Pitcock <nenolod@nenolod.net>",
+	"  spb, Stephen Bennett <spb@exherbo.org>",
+	" ",
+
+	"The following additional people have contributed to Atheme:",
+	" ",
+	"  Adam, Adam <adam@sigterm.info>",
+	"  aji, Alex Iadicicco <alex@ajitek.net>",
+	"  alambert, Alex Lambert <alambert@quickfire.org>",
+	"  amdj, Aaron M. D. Jones <aaronmdjones@gmail.com>",
```
ilbelkyr: I'll be honest, given the amount of work you've done for Atheme you should probably be listed at least in the group above, if not the core group (cc @maxteufel, @siniStar7boy about that one, though).

libathemecore/version.c
aaronmdjones: Fixed in commit 7bab1e5e718faa9bc237c493fdb7cb2f652b5492

libathemecore/version.c
siniStar7boy: I would agree on this completely and was actually thinking the exact same thing.

PR #644, opened 2018-11-17
translated to turkish

Luffy:
[tr.tar.gz](https://github.com/atheme/atheme/files/2591567/tr.tar.gz)


aaronmdjones:
I will integrate the files from the tarball you attached.

PR #658, opened 2019-05-16
Update SASL-SCRAM-SHA document

mquin:
Correct and clarify note on digest selection.


ilbelkyr:
Closing per 5783710fe as we've decided to adjust that part a little more, but thank you for pointing this out!

PR #660, opened 2019-05-24
Updated nickserv/info to show vhostnick vhosts

alyx:
**HostServ/VHOSTNICK** applies per-nick vhosts using an alternative metadata format `private:userhost:<nick>`. Currently very few, if any, modules within other services built to process vhosts will show per-nick vhosts due to this.

This patch updates **NickServ/INFO** to correctly display all virtual hosts attached to an account, including both the account-based host and all nick-based hosts.


alyx:
Resolves #661

Issue #676, opened 2019-10-26
Rework AKICKs


ilbelkyr:
Currently, akicks are a special kind of `chanacs` entry with some metadata set. This makes for all sorts of awkward special cases. We should make them a separate thing.

PR #677, opened 2019-10-26
entity validation: avoid indirection via chanacs

ilbelkyr:
Poking @aaronmdjones since this might affect some of the stuff you've been working on, plus for an overall review if possible.

This will help with #676, which is my primary motivation for getting this in for 7.3.0.


ilbelkyr:
Merged per IRC discussion.

PR #680, opened 2019-12-06
Remove the last remaining '_t' data types (from the hook system & modules/scripting/perl/)

aaronmdjones:
Types with names ending in `_t` are reserved and should not be defined by applications.

This also moves all data structures passed to hook functions to one place, `include/atheme/hook.h`, and they are now sorted alphabetically, like in `include/atheme/hooktypes.in`. This makes finding/adding/deleting/modifying hooks easier.

Finally, a few confusing or inconsistent hook data structures were renamed. The hooks themselves were not renamed, but this may still affect some out-of-tree modules. The most obvious one is `hook_cmessage_data_t` -> `struct hook_channel_message`.

### DIFF
--- a/include/atheme/hook.h
+++ b/include/atheme/hook.h
@@ -81,6 +81,17 @@ struct hook_channel_topic_check
 	int             approved;       // Write non-zero here to cancel the change
 };
 
+struct hook_expiry_req
+{
+	union {
+		struct mychan * mc;
+		struct myuser * mu;
+		struct mynick * mn;
+	}                       data;
+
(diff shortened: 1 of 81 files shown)


include/atheme/hook.h
```diff
@@ -104,6 +104,14 @@ struct hook_expiry_req
 	int                     do_expire;  // Write zero here to disallow expiry
 };
 
+struct hook_host_request
+{
+        const char *        host;
```
ilbelkyr: Looks like the indentation is off on this one (though it's been that way before the move as well); we might as well take the opportunity to fix it

PR #685, opened 2019-12-30
modules/crypto/argon2: fix compilation without libargon2

dwfreed:
If libargon2 isn't available, then CRYPTO_MODULE_NAME wasn't being defined for the SIMPLE_DECLARE_MODULE_V1() at the end, so I moved it outside the ifdef.


aaronmdjones:
Oops. My bad; and CI didn't catch it because I told it to install the library ...

PR #690, opened 2020-02-08
libathemecore: rip out C99-snprintf compat code

aaronmdjones:
We now rely heavily on a proper C99 compiler for the following:

- C++-style single-line comments:

```C
    // This is a C99 comment that does not need a closing element
    // These comments continue only to the end of the line they are on
```

- Variable declaration in for:

```C
    for (unsigned int i = 0; i < 4; i++)
```

- Designated initialisers:

```C
    struct foo {
        unsigned int a;
        unsigned int b;
    };

    static struct foo bar = {
        .a = 4,
        .b = 8,
    };
```

- Variable declarations anywhere in a function:

```C
    void function(void)
    {
        (void) fprintf(stderr, "Hello, world!\n");

        const unsigned int answer = 42;

        (void) fprintf(stderr, "The answer to life, the universe, and "
                               "everything, is: %u\n", answer);
    }
```

- Block-scope variable declarations:

```C
    void function(void)
    {
        if (some_condition)
        {
            const unsigned int x = 4;
            (void) fprintf(stderr, "%u\n", x);
        }

        // x does not exist at this point
    }
```

- The static qualifier in a function parameter list:

```C
    void foo(unsigned int arr[static 4])
    {
        // Would explode if passed an array with fewer than 4 elements.
        for (unsigned int i = 0; i < 4; i++)
            (void) fprintf(stderr, "%u ", arr[i]);

        (void) fprintf(stderr, "\n");
    }

    void bar(unsigned int num[static 1])
    {
        /* Decays to "unsigned int *num" but also indicates that a NULL
         * argument is not allowed. This would explode if passed NULL.
         */
        (void) fprintf(stderr, "%u\n", num[0]);
    }

    void baz(void)
    {
        unsigned int arr[] = { 1, 2, 3 };

        // Generates a compile-time error; 'arr' is not large enough:
        (void) foo(arr);

        // Likewise:
        (void) bar(NULL);
    }
```

- The restrict qualifier in a function parameter list:

```C
    void foo(unsigned int *restrict a, unsigned int *restrict b)
    {
        // Compiler is now free to re-order these two stores:
        *a = 8;
        *b = 4;

        /* Saves emitting at least one load instruction.
         * (Does not have to reload *a after assigning to *b)
         * May always print 8 even if both pointers are the same!
         */
        (void) fprintf(stderr, "%u\n", *a);
    }
```

- Anonymous variables in function calls:

```C
    struct foo {
        unsigned int a;
        unsigned int b;
    };

    void bar(const struct foo *const foo)
    {
        (void) fprintf(stderr, "%u %u\n", foo->a, foo->b);
    }

    void baz(void)
    {
        // Prints 4 8
        (void) bar((&(struct foo){ 4, 8 }));
    }
```

Since this is now the case, a non-C99 compiler cannot build us anymore. We therefore do not need the `libathemecore/snprintf.c` fallback file.

Additionally, that file has many shortcomings:

- If you are cross-compiling, it is always compiled in and used because the m4 file does a run-time test to see if it is necessary and the test (sensibly, I might add) always fails during a cross-compile.

- Due to the above, compiler optimisations (such as constant literal propagation) and security features (such as source fortification) are not applied to any `(v)snprintf(3)` calls if we are being cross-compiled.

- This implementation does not support positional format tokens (`%n$s`), which are necessary for translations.

Rip out the fallback implementation. Also confirm whether positional format tokens work (which is not guaranteed), and disable localization support if they don't.


libathemecore/atheme.c
```diff
@@ -183,6 +183,94 @@ detach_console(int *daemonize_pipe)
...
+		return false;
+
+	if (snprintf(buf, 0U, "No, it's %32zu.", fourty_two_sz) != 42)
+		return false;
+
+	if (strncmp(buf, "The answer is", 14U) != 0)
+		return false;
+
+	return true;
+}
```
ilbelkyr: 1. Nice example text, I approve
2. Apologies if you already mentioned this internally; could we test for this at configure time instead? (That wouldn't catch cross-compilation troubles, though I suspect we already fail at that. Also, "yes but autoconf" is probably an acceptable answer.)

alyx:
@aaronmdjones have you had the opportunity to evaluate if this is a safe change or not?

libathemecore/atheme.c
aaronmdjones: No need to test at configure or runtime now that the configure script itself aborts if we're not using a C99 compiler. This test has been removed.

aaronmdjones:
When correcting for the bogus `install-sh` failure in the build system (I'll get to that some other time), I have confirmed that this works on at least OpenBSD 6.8 and FreeBSD 12.2, including positional format token support (so localization is possible).

ilbelkyr (approved):

PR #691, opened 2020-02-13
Stop using "negative" channel access flags for AKICKs

ilbelkyr:
There's still work to be done on this particularly to create a migration path for chanacs-based akicks, hence this is not ready for merge yet. I'm creating this as a place for reviews already, though.

### DIFF
--- a/modules/chanserv/akick.c
+++ b/modules/chanserv/akick.c
@@ -9,62 +9,425 @@
 
 #include <atheme.h>
 
-struct akick_timeout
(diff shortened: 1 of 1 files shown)


kaniini:
when I designed the chanacs system to use +b for akicks it was intentional.  please do not change a fundamental design element of atheme.

ilbelkyr:
(don't bother reviewing this for now. #676 is probably a better place to discuss this vs alternative approaches)

PR #693, opened 2020-04-09
libathemecore: handle CTCPs case-insensitively

dwfreed:
Switch CTCP handling to case-insensitive to match modern IRC docs

### DIFF
--- a/libathemecore/ctcp-common.c
+++ b/libathemecore/ctcp-common.c
@@ -67,7 +67,7 @@ ctcp_about_handler(struct sourceinfo *si, char *cmd, char *args)
 void
 common_ctcp_init(void)
 {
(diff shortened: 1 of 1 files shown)


libathemecore/ctcp-common.c
```diff
@@ -67,7 +67,7 @@ ctcp_about_handler(struct sourceinfo *si, char *cmd, char *args)
 void
 common_ctcp_init(void)
 {
-	ctcptree = mowgli_patricia_create(noopcanon);
+	ctcptree = mowgli_patricia_create(irccasecanon	);
```
aaronmdjones: This should use `strcasecanon` in case the configured match mapping is not ASCII.

alyx:
Closed by #732

PR #694, opened 2020-04-12
nickserv/info: Make private last-seen times more private

edk0:
The existing approach rounds the elapsed time down to weeks, which does reduce the resolution of last seen times to a week over one observation. Unfortunately, if we're allowed multiple observations, we can zero in on the time the low-resolution clock ticks over.

The solution used here is to reduce the resolution of the last-seen time itself. Since the time is not changing, this happens the same way each time, and subsequent observations reveal no information. This is not perfect on its own: it discloses high-resolution information about users who were last seen at the beginning of the current Atheme week, since they couldn't have been last seen in the future. We deal with this by representing both 0- and 1-week differences as "less than two weeks ago".

Finally, this is of limited value as long as we reveal the high-resolution fact that a user is online now. We dispense with this and show an obfuscated time for private users regardless of logged-in status.

### DIFF
--- a/modules/nickserv/info.c
+++ b/modules/nickserv/info.c
@@ -11,6 +11,18 @@
 
 static bool show_custom_metadata = true;
 
+static const char *
+ns_obfuscate_time_ago(time_t time)
+{
+	time_t diff = CURRTIME / SECONDS_PER_WEEK - time / SECONDS_PER_WEEK;
+	static char buf[BUFSIZE];
+
+	if (diff <= 1)
+		return _("(less than two weeks ago)");
+	snprintf(buf, ARRAY_SIZE(buf), _("(about %lu weeks ago)"), (unsigned long)diff);
+	return buf;
+}
+
 static void
 ns_cmd_info(struct sourceinfo *si, int parc, char *parv[])
 {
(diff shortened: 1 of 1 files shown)


modules/nickserv/info.c
```diff
@@ -198,40 +210,40 @@ ns_cmd_info(struct sourceinfo *si, int parc, char *parv[])
...
 		strftime(lastlogin, sizeof lastlogin, TIME_FORMAT, tm2);
 		if (hide_info)
-			command_success_nodata(si, _("Last seen  : (about %u week(s) ago)"), (unsigned int)((CURRTIME - mn->lastseen) / SECONDS_PER_WEEK));
+			command_success_nodata(si, _("Last seen  : %s"), ns_obfuscate_time_ago(mn->lastseen));
 		else
 			command_success_nodata(si, _("Last seen  : %s (%s ago)"), lastlogin, time_ago(mn->lastseen));
 	}
 
 	/* if noone is logged in to this account, show account's lastseen,
 	 * unless we have a nick and it quit at the same time as the account
```
ilbelkyr, line 230: Minor nitpick, but could you add a line to the comment to make it clear that if we `hide_info`, we always act as though no one was logged in to avoid leaking that info?

PR #711, opened 2020-07-26
Do not prevent users from using regain while +b/+q

ddevault:
+q in particular is often used to cast a wide net, with masks like +q
$~a, which causes normal users to be unable to use regain without
parting as many as tens of channels first. The prohibition was never
useful in the first place and is causing problems now, so it's gotta go.

### DIFF
--- a/modules/nickserv/enforce.c
+++ b/modules/nickserv/enforce.c
@@ -445,13 +445,6 @@ ns_cmd_regain(struct sourceinfo *si, int parc, char *parv[])
 	}
 	if ((si->smu == mn->owner) || verify_password(mn->owner, password))
 	{
-		struct chanuser *cu = NULL;
-		if (si->su != NULL && ((cu = find_user_banned_channel(si->su, 'b')) || (cu = find_user_banned_channel(si->su, 'q'))))
-		{
-			command_fail(si, fault_noprivs, _("You can not regain your nickname while banned or quieted on a channel: \2%s\2"), cu->chan->name);
-			return;
-		}
-
 		if (qline_find_match(target) && !has_priv(si, PRIV_MASS_AKILL))
 		{
 			command_fail(si, fault_noprivs, _("You can not regain a reserved nickname."));


coderobe:
This is the same behavior as 435 BANNICKCHANGE as implemented by some IRCds, dropping this behavior *without* also merging something like https://github.com/charybdis-ircd/charybdis/pull/354 would allow bypassing 435 by regaining. Given that, this should be configurable behavior IMO

ddevault:
This is the Wrong Thing To Do, and 435 should be buried and forgotten. Making it configurable would be implementing the bugfix incorrectly. I will prepare corresponding patches for each supported ircd, but I will not make this configurable.

ilbelkyr:
Services deliberately match ircd behaviour in this case.

If you disagree with the corresponding ircd functionality, please take this up with the ircd developers. Should ircds behave differently in this regard, services would obviously have to adjust to the ircd behaviour appropriately there.

ddevault:
I have, and they told me to patch Atheme as well. Don't send me in circles.

ddevault:
unrealircd patch: https://github.com/unrealircd/unrealircd/pull/118

ilbelkyr:
Unless every supported ircd changes its behaviour unconditionally in all supported versions, this PR does not appropriately address the matter as it would also apply to ircds or configurations thereof which still exhibit the behaviour this code mimics.

In any case, services reflect ircd behaviour here. I'd rather not apply such a change to services before ircd developers express the intent to proceed with this.

kaniini:
As the person who came up with `REGAIN` and `+q $~a` I actually agree with @ddevault.

kaniini:
Here is my justification as for why: REGAIN logs you into your account as a side effect.  Therefore, this is not the same scenario as 435, as you *are* logged in.  That the IRCd does not yet see that side effect is not a problem.

ddevault:
The main problem with IRC has nothing to do with the shiny features IRCv3 attempts to address, and everything to do with the old boy's club of grumpy maintainers and admins who will shut down (or ignore) any changes, no matter how trivial or correct, which haven't been signed in triplicate by at least 5 people that they've been known for at least 15 years.

You are the cancer which is killing IRC.

ilbelkyr:
For the specific situation of `+q $~a`, it's currently quite awkward indeed with REGAIN indeed logging you in. I thought we had at some point changed it to log you in *before* checking for bans for this specific reason, though it looks like that's not the case

ddevault:
Disruptive content? Damn straight it is, and your broken system needs disrupting. It has needed it for a long, long time.

kaniini:
@ddevault 

I don't think that comment is necessary.  This is just a matter of nuance, and we can work together to come to a conclusion that works for everyone.  Trying to come in fists swinging isn't going to work with @ilbelkyr anyway, she stands her ground, especially when she believes she is right.  Trust me, I know from experience.

@ilbelkyr 

Since we agree that REGAIN does cause login as side-effect, I think we should drop this silly restriction.  It is better to bend the rules and provide a good user experience instead of giving people a crap experience on technicality.

ddevault:
Just keep calm and carry on, like you always have, nothing is wrong, IRC is perfect, and anyone who suggests otherwise is to be silenced. Just draw those curtains so you don't have to watch the bleeding husk of the IRC community atrophying users and you can keep living as if everything is okay, forever and ever, until you're the last one left. You'll have to ban me to shut me up, I want you to explicitly and deliberately silence the opposition, as a direct gesture that the status quo is to be upheld.

aaronmdjones:
The correct fix would be to simply log you in before checking for bans, because it's possible that doing so is going to result in you no longer being banned.

Your comment was marked as disruptive because you are attacking the person rather than the idea, and yes, if you do continue to engage like that, the conversation will be locked.

If you're not willing to write the patch described above, I will do so when I have some free time.

kaniini:
Thanks @aaronmdjones!  Can we get a backport to 7.2 so that it lands on freenode quickly?

ddevault:
The correct change is to remove this "feature". Code has to be justified to exist, it isn't valid simply because it's there. If you can come up for a legitimate reason for this feature to exist, then your suggested fix would be correct.

kaniini:
The feature has relevance, for example with accounts that are restricted.  Applying the login side effect early fixes the bug while ensuring that restricted accounts still cannot use REGAIN to bypass nick change restrictions.

ddevault:
Nick change restrictions are not a valid feature in the first place. It's none of the channel operator's business what your global nick is, it's beyond their domain of privilege. If the user's nick changing is disruptive, the user should be removed, the same for any other disruptive behavior.

kaniini:
Nick change restrictions are important.  Otherwise you get people evading the `+q` by changing their nickname in response to being `+q`.

ddevault:
No, because such +q's (and +b's) are trivially circumvented by parting, changing your nick, and rejoining. Nick change restrictions have never been an effective moderation tactic. Only naieve operators ban or quiet users by nick and expect it to stick.

kaniini:
Nick change restrictions match any kind of mask, not just nicks...

ddevault:
Correct, and such as sufficiently specific mask would not be affected by a nick change to circumvent a +q. You're mixing unrelated things: nick change restrictions, and quiets. The former is not an effective tactic to combat circumvention of the latter.

kaniini:
It has worked well enough for *checks notes* the almost 16 years since I first implemented it in hyperion.

ddevault:
Maybe you don't understand.

```
<aggressor> flame flame flame
* operator sets mode +b aggressor!*@*
* aggressor parts #channel 
* aggressor2 joins #channel 
<aggressor2> flame flame flame
```

kaniini:
Yeah, that's an operator who does not have sufficient knowledge to effectively moderate their channel.  What the nick change restriction is there to prevent is children using /nick to continue their temper tantrum.

Observe:

```
<idiot> flame flame flame
* op sets mode +q *!*@*.aol.com
* idiot changes nick to fuck
* idiot changes nick to you
* idiot changes nick to asshole
```

With nick change restriction this is not possible.

ddevault:
In the situation you described, it should obviously be escalated from +q to +b.

kaniini:
The point of features like these are to prevent the need for escalation in the first place.

This is a key part of the philosophy behind the design and implementation of the atheme+charybdis stack in the first place, building de-escalation into the software.  Similarly, we went to a lot of effort to automate tasks normally done by the IRCop in traditional IRC, so that they could spend their time empathizing with users to de-escalate tantrums.  The latter has not worked out fully as intended, but I believe the work to have been a net positive.

If you want "old-school" IRC, ircd-hybrid 6 can still be downloaded to this day.

ddevault:
+b is an escalation in that it applies stronger sanctions, but it need not be an escalation in that it's some kind of harsher punishment. It's applying the correct tool to the job: the user is being a brat and +q isn't sufficient, so +b is used. The operator may choose to remove it later, just as they would a +q. Solving social problems with technology has never worked well, and in this case, it has unintended side effects on innocent users.

kaniini:
I think you misunderstand what `+q` is actually for.

Unfortunately, the person who first conceptualized `+q` is no longer with us, so you'll have to take my word on this, but since it was first conceived, the purpose of `+q` is to deny write access to a channel.  That means any kind of write to the channel, including a nick change.

Anyway, I think this discussion has outlived its usefulness.

ddevault:
Nick changing isn't a write to the channel, it's a write to the network. They can part and rejoin, too, which is a "write" to the channel by the same logic.

kaniini:
*sigh* it is both a write to the network and a write to the channel as the `NICK` message is propagated to all common channel members.

ddevault:
So is PART and JOIN.

kaniini:
yes, and `PART` and `JOIN` have to be exceptions to that policy for obvious reasons, now don't they?

arguing about this is a waste of my time, bye

ddevault:
NICK seems to have the same obvious objections if you ask me. I agree that this is a waste of time.

PR #712, opened 2020-07-26
NickServ REGAIN: log in before changing nick

ddevault:
This addresses a race condition where the user is "banned" on a channel
because all unregistered users are banned, preventing them from changing
their nick.

---

This is an alternate take on #711. I maintain that this is the wrong fix, but it's your project, so here you go.


kaniini:
looks good to me

dwfreed:
The maxlogins check needs to be moved as well, I think?

alyx:
looks good to me, sorry for this being delayed.

ddevault:
Thanks.

PR #716, opened 2020-08-03
ophion protocol: core support

kaniini:
This PR implements the bare minimum required for atheme to link to an Ophion network and act as an identity service.

From a S2S protocol perspective, Ophion is a transactional variant of TS6 with support for distributed channel ACL, which can make use of several kinds of IRC-facing and web-facing services.

An identity service provides authentication, authorization and integration with external databases such as LDAP, while a persistence service backs up the distributed state to permanent storage.

Future identity-related work will involve pushing group membership to users in addition to their account names, allowing for Ophion ACLs to match against group memberships.

Allowing atheme to run as an Ophion persistence service is also planned.  Finally, support for the Ophion webhook service may also be implemented in atheme down the line.  This allows for third-party services to integrate into the Ophion network using webhooks.

### DIFF
--- a/dist/atheme.conf.ophion-identity-example
+++ b/dist/atheme.conf.ophion-identity-example
@@ -1,0 +1,1493 @@
+/* This is an example configuration for running an Ophion identity
+ * service using Atheme.  In most cases, you will want to use
+ * atheme.conf.example instead.  This configuration file should be
+ * saved as atheme.conf.
+ *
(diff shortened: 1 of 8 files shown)


kaniini:
any reason this has not been reviewed?

alyx:
Per our discussion the other week, this is no longer something valuable to ophion; closing this PR. 
If that's changed and services support is useful, feel free to reach out.

Issue #652, opened 2019-03-07
Feature Request: maxlogins per account


cooldude13233:
I hope that some dev implements some module of specifying maxlogins per account instead of having to justify maxlogins on accounts network wide

alyx:
Can you clarify the benefit of this feature?

cooldude13233:
@alyx it’s so that accounts are to have flexibility to link other nicknames together instead of being constrained by the global limit specified by default

jesopo:
the usecase I care about here is things like services bots that all do a similar thing but are split in to separate instances, for whatever reason.

I think there's two options here:

1. have two categories for `OS SET MAXLOGINS` - a low and a high limit, where a flag to use the high limit is set on accounts
2. a privileged `NS SET MAXLOGINS` that, without being set, defaults to `me.maxlogins`

any better ideas?

alyx:
Looking at it from an oper perspective makes sense. The one thing I'd wonder is, if we're going in from an OS perspective anyway, would it be better to have a high limit (and thus, still have a limit that will somehow end up being hit again) or just bump it to "unlimited logins"?

jesopo:
how about something like `NS LOGINLIMIT <nick> ON|OFF` or so?

has been discussed that an FSET with privilege control over each possible setting would be nice but that's a bigger discussion

ilbelkyr:
fixed with #723, I'd say

PR #723, opened 2020-11-03
implement NS LOGINNOLIMIT <nick> ON|OFF

jesopo:
as per #652. most of this was stolen from how regnolimit works.

n.b. I do not speak very good C


alyx:
This seems fine, thank you for your contribution. Apologies about the delay in review.

Issue #728, opened 2020-11-09
avoiding leaking oper status in remote whois


jesopo:
some networks are configured to not show oper status in whois and some ircds have differing umodes (think +o/+O) for different levels of privilege, and atheme remote whois unconfigurably giving `RPL_WHOISOPERATOR` ends up leaking information

alyx:
Reposting here for visibility.

Per developer discussion, we agree that it may not always be ideal to leak oper status during a remote whois, however we feel the best course of action moving forward is a configurable option to strip the `RPL_WHOISOPERATOR` oper string from a remote whois.

jesopo:
do we want a new `atheme.conf` bool for this or do we want to treat `operstring = "";` (empty string) as "don't send RPL_WHOISOPERATOR"

alyx:
I think an additional config bool would make the most sense. @ilbelkyr any thoughts?

alyx:
Closed by #733

PR #730, opened 2020-11-14
refuse to answer remote WHOIS for users not on our server

jesopo:
I have a strong feeling this isn't going to be an acceptable solution to #728 but I can't actually think of a good reason for Atheme to tolerate remote whoises for remote users

### DIFF
--- a/libathemecore/ptasks.c
+++ b/libathemecore/ptasks.c
@@ -351,7 +351,7 @@ handle_whois(struct user *u, const char *target)
 	if (floodcheck(u, NULL))
 		return;
 
-	if (t != NULL)
+	if (t != NULL && t->server == me.me)
 	{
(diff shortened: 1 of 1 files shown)


alyx:
Hi there,

Thank you for the patch, this is definitely a possible solution to this issue and we really appreciate the patch :) 

That said, per discussion with the other developers, we feel this isn't the best solution. It would be unusual for an IRCd (which Services pretends to be) to not respond with any answers for a remote whois; ideally we think if hiding oper status is to be implemented, the correct path should be via hiding specifically the RPL_WHOISOPERATOR string based on a configurable option set in atheme.conf.

If you're interested in contributing this change, feel free to resubmit a new PR when you feel like it! Otherwise, we're happy to take this into consideration moving forward during our own work on Atheme.

Thanks again, feel free to reach out with any questions you may have!

PR #735, opened 2020-12-03
Inspircd: improvements to the module's cmode stuff

Elizafox:
More extban types are added, a lot of mode parameter stuff was cleaned up, and support for some contrib modules was added too. In addition, durations are now supported for m_chanhistory.

Still to do: make it work with protocol 1205. But that's outside the scope of the work I've done so far.

### DIFF
--- a/modules/protocol/inspircd.c
+++ b/modules/protocol/inspircd.c
@@ -10,6 +10,7 @@
 #include <atheme.h>
 #include <atheme/protocol/inspircd.h>
 
+// TODO: protocol 1204
 #define PROTOCOL_MINIMUM 1202 // we do not support anything older than this
 #define PROTOCOL_PREFERRED_STR "1202"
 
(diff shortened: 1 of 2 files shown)


modules/protocol/inspircd.c
```diff
@@ -10,6 +10,7 @@
 #include <atheme.h>
 #include <atheme/protocol/inspircd.h>
 
+// TODO: protocol 1204
```
SadieCat, line 13: Protocols 1203 and 1204 never actually shipped and InspIRCd will not link against them. You can just support 1202 (v2) and 1205 (v3).

Elizafox: I typo'd and meant 1205 lol

modules/protocol/inspircd.c
```diff
@@ -107,18 +112,62 @@ static bool has_svshold = false;
 static bool has_cloakingmod = false;
 static bool has_shun = false;
 static bool has_svstopic_topiclock = false;
+static bool has_bannegate = false;	// Contrib module. --Elizafox
 static int has_protocol = 0;
 
+/* NOTE: modifies argp and argl, but not str
+ * You must have allocated at least maxsize space in argp and argl!
+ */
```
ilbelkyr, line 120: Might be worth adding a high-level summary of the function (including inputs and outputs) for clarity

modules/protocol/inspircd.c
```diff
@@ -134,18 +183,21 @@ inspircd_next_matching_ban(struct channel *c, struct user *u, int type, mowgli_n
 		if (cb->type != type)
 			continue;
 
-		if ((!match(cb->mask, hostbuf) || !match(cb->mask, realbuf) || !match(cb->mask, ipbuf)) || !match_cidr(cb->mask, ipbuf))
-			return n;
+		if ((!match(mask, hostbuf) || !match(mask, realbuf) || !match(mask, ipbuf)) || !match_cidr(mask, ipbuf))
+			return invert ? NULL : n;
```
ilbelkyr, line 188: hrm, correct me if I'm misunderstanding things – it looks like you're checking for a normal match, exiting early if bannegate was used on the banmask. It seems like we'd need to invert the matching logic instead (though note that if a ban doesn't match against a user's `nick!user@host`, that'll only be an inverted match if it's not an extban instead)

modules/protocol/inspircd.c
```diff
@@ -162,20 +231,27 @@ inspircd_next_matching_ban(struct channel *c, struct user *u, int type, mowgli_n
...
+			case 'z':
+				matched = !match(p, u->certfp);
 				break;
 			default:
 				continue;
 			}
 
 			if (matched)
-				return n;
+				return invert ? NULL : n;
```
ilbelkyr, line 255: (as above)

modules/protocol/inspircd.c
```diff
@@ -208,30 +284,46 @@ channel_metadata_sts(struct channel *c, const char *key, const char *value)
...
+	const char *p = value;
+	long long ret;
 
-	p = value;
-	arg2 = NULL;
+	if(len < 0 || len > 10)
+		return -1;
 
-	while (*p != '\0')
+	for( ; *p != '\0' || (len && (value - p) > len); p++)
```
ilbelkyr, line 300: Seems like a good idea to move `const char *p = value` into the first part of the for loop head here – this both avoids an empty for loop initial step and moves the declaration to where it's needed

modules/protocol/inspircd.c
```diff
@@ -250,7 +342,32 @@ check_nickflood(const char *value, struct channel *c, struct mychan *mc, struct
...
+	if(argl[0] == 0 || argl[0] > 10 || argl[1] == 0 || argl[1] > 10)
+		return false;
+
+	if(param_int(argp[0], argl[1]) <= 0)
+		return false;
+
+	/* History can be a duration or integer
+	 * XXX - find out maximum history duration and check it
+	 */
+	for(p = argp[1] ; *p != '\0'; p++)
```
ilbelkyr, line 362: Also a good idea to move the declaration of `const char *p` here

modules/protocol/inspircd.c
```diff
@@ -273,55 +390,90 @@ check_forward(const char *value, struct channel *c, struct mychan *mc, struct us
...
-		if (!isdigit((unsigned char)*ch))
-			return false;
-		ch++;
-	}
+	if(param_int(value, 0) <= 0)
+		return false;
 
-	if (atoi(value) <= 0 || atoi(value) > (int) max_rejoindelay)
-	{
+	delay = atoi(value);
```
ilbelkyr, line 399: hrm, could this use the return value from `param_int` instead?

ilbelkyr:
@Elizafox apologies for the delay – we do appreciate the contribution, but as usual, life got in the way for a while :(

There's a few things I've commented on that seem like they might need to be fixed, as well as some minor style/cleanliness improvements.

I'm insufficiently familiar with InspIRCd to judge that part of this PR, sadly – however I've spoken with @SadieCat about this a bit out of band and I think generally this should be fine for now, though we should eventually support the updated protocol. That said, as long as it's still supported, updating the older protocol is better than leaving it all to sit and go nowhere.

As for future directions, there's a fair bit of configurability in inspircd modes as well as its behaviour which we aren't currently equipped to handle in all its potential variations. I suspect we'll want to look into this at some point and see how much of that can be addressed via some creative protocol module code. (Perhaps that'll be an opportunity to review some assumptions we make about protocols and revise the protocol module API to allow for certain things to be determined at startup/link time if there's a way to do so without taking apart the entirety of services, too.)

PR #739, opened 2020-12-06
RETURN to a different address should HIDEMAIL if in defuflags

jesopo:
per #737

if:
- user is not set HIDEMAIL and
- HIDEMAIL in defuflags and
- new email does not equal old email

then force HIDEMAIL on the user


ilbelkyr:
it may be a good idea to send an additional response explicitly noting that the `HIDEMAIL` setting has been changed, I think

Issue #743, opened 2021-01-20
priv to give immunity from being kicked in a closed channel


jesopo:

jesopo:
maybe it would just make sense to not kick (s)opers from closed channels, it doesn't really feel like something that needs fine-grained control like a priv

jesopo:
perhaps chanserv could just avoid kicking you if it's already joined the channel to set the closed modes. if you're able to get in after that, chances are you're probably override and shouldn't be kicked

ilbelkyr:
> perhaps chanserv could just avoid kicking you if it's already joined the channel to set the closed modes. if you're able to get in after that, chances are you're probably override and shouldn't be kicked

(looks like that's what we went with in #744)

jillest:
Why does anyone need to be in a closed channel? Since it is locked down on ircd level, it is not like you can use this to kline bots that attempt to join.

jesopo:
some IRCd don't let you change the topic, even with oper overrides, if you're not in the channel.

PR #744, opened 2021-01-20
use try_kick() instead of kick() for channels that have been CS CLOSEd

jesopo:
my proposed solution for #743, stolen from cs/akick.c, still sets the closed channel modes but announces that it isn't kicking you because you're oper/immune.

I did consider instead only kicking users that joined when the channel is already closed, because anyone that can join through `+b *!*@*` should probably be allowed to, but that means you need to do join/kick before you can get in safely, and it's likely a bit vulnerable to timing attacks.


ilbelkyr:
mm, there's another `kick()` further down the code as part of initially clearing the channel when the command is first used; i suppose for consistency, this should use `try_kick()` as well?

aaronmdjones (approved):

Issue #748, opened 2021-02-02
support !T duration format in OS CLEARCHAN AKILL


jesopo:
also maybe make it default to `config_options.kline_time` rather than a hardcoded week

ilbelkyr:
@aaronmdjones i think you mentioned you wanted to look into this, i don't remember with quite enough certainty to just assign the issue to you though – feel free to self-assign or otherwise let me know and I'll see about looking into it :)

PR #749, opened 2021-02-03
config_options.kline_time for CLEARCHAN AKILL, not a hardcoded week

jesopo:
mentioned on #748


jesopo:
if we get `!T` parsing refactored out somewhere I can use it, I'll put that change in this PR too

jesopo:
need to update `CLEARCHAN` help files too

PR #751, opened 2021-02-08
chanserv: allow hiding AKICK entries from public ACLs

ilbelkyr:
This essentially allows networks to enable PUBACL even if they can't make akicks public (e.g. networks like freenode that already make ACLs public but exclude akicks from the public listing, so exposing akicks isn't an option without breaking users' existing assumptions about them being private).

Tested, but there's likely to be some edge cases I haven't thought of.

### DIFF
--- a/modules/chanserv/count.c
+++ b/modules/chanserv/count.c
@@ -58,6 +58,11 @@ cs_cmd_count(struct sourceinfo *si, int parc, char *parv[])
 	ca_hop = get_template_flags(mc, "HOP");
 	ca_vop = get_template_flags(mc, "VOP");
 
+	bool show_akicks = true;
+
+	if (chansvs.hide_pubacl_akicks)
+		show_akicks = ( chanacs_source_has_flag(mc, si, CA_ACLVIEW) || has_priv(si, PRIV_CHAN_AUSPEX) );
+
 	MOWGLI_ITER_FOREACH(n, mc->chanacs.head)
 	{
 		ca = (struct chanacs *)n->data;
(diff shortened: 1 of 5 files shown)


modules/chanserv/count.c
```diff
@@ -75,17 +80,31 @@ cs_cmd_count(struct sourceinfo *si, int parc, char *parv[])
...
+		else
+			command_success_nodata(si, _("%s: VOP: %u, HOP: %u, AOP: %u, SOP: %u, Other: %u"),
+					chan, vopcnt, hopcnt, aopcnt, sopcnt, othercnt);
+	}
 	snprintf(str, sizeof str, "%s: ", chan);
 	for (i = 0; i < ARRAY_SIZE(chanacs_flags); i++)
 	{
 		if (!(ca_all & chanacs_flags[i].value))
 			continue;
+		if (chanacs_flags[i].value == CA_AKICK && !show_akicks)
```
aaronmdjones, line 106: What about entries that have `+b` *and* any other flag (except `+e`)? Those are still enforced as akicks, but would be listed here. I suppose that would be a valid way for channel operators to denote which akicks are supposed to be public (e.g. `+Ab`), but it seems a bit icky.

This is because the way akicks in general are implemented is less than ideal, but we already knew that much.

modules/chanserv/flags.c
```diff
@@ -317,7 +331,7 @@ cs_cmd_flags(struct sourceinfo *si, int parc, char *parv[])
 				target = sstrdup(mt->name);
 				ca = chanacs_find_literal(mc, mt, 0);
 			}
-			if (ca != NULL)
+			if (ca != NULL && (ca->level != CA_AKICK || show_akicks))
```
aaronmdjones, line 334: Likewise.

modules/chanserv/flags.c
```diff
@@ -115,6 +120,9 @@ do_list(struct sourceinfo *si, struct mychan *mc, unsigned int flags)
 
 		ca = n->data;
 
+		if (ca->level == CA_AKICK && !show_akicks)
```
aaronmdjones, line 123: Likewise.

modules/chanserv/count.c
ilbelkyr: There's some precedent for this in `chanserv/access`, which already skips over entries that are plain `+b` while handling other entires with `+b` normally. (That said, you could argue that it should also process the entry as though it didn't have the `+b` – though that'll bring a number of complications too, so that's perhaps a question for another time.)

Rethinking it, I'd tend to agree with you, though.

> (except `+e`)

Not sure `+be` should be treated specially any more than, say, `+bo` for this purpose; we do show entries with just plain `+e`, after all. (There might be an argument to be made for them to be private, analogous to ircd-side `+e`, but I haven't seen any demand for that.)

aaronmdjones: To be clear, I wasn't arguing for `+be` to be treated specially; rather, pointing out that a `+be` entry isn't enforced.

If there's already precedent for this behaviour I suppose that's fine.

aaronmdjones (approved):

Issue #753, opened 2021-02-09
Non-sopers can view private metadata via xmlrpc api


examknow:
Anyone can view whatever metadata about accounts or channels they want without logging in as a soper using the atheme.metadata method via the xmlrpc api.

PR #754, opened 2021-02-09
prevent non-sopers from viewing private metadata via xmlrpc api

examknow:
See #753

### DIFF
--- a/modules/transport/xmlrpc/main.c
+++ b/modules/transport/xmlrpc/main.c
@@ -488,6 +488,8 @@ xmlrpcmethod_ison(void *conn, int parc, char *parv[])
 /* atheme.metadata
  *
  * XML inputs:
+ *	 authcookie
+ *       account name
  *       entity name, UID or channel name
  *       metadata key
  *       metadata value (optional)
@@ -499,6 +501,7 @@ static int
 xmlrpcmethod_metadata(void *conn, int parc, char *parv[])
 {
 	struct metadata *md;
+	struct myuser *mu;
 	int i;
 	char buf[XMLRPC_BUFSIZE];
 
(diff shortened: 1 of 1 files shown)


modules/transport/xmlrpc/main.c
```diff
@@ -488,6 +488,8 @@ xmlrpcmethod_ison(void *conn, int parc, char *parv[])
 /* atheme.metadata
  *
  * XML inputs:
+ *	 authcookie
```
aaronmdjones, line 491: This is indented incorrectly.

modules/transport/xmlrpc/main.c
```diff
@@ -511,49 +514,68 @@ xmlrpcmethod_metadata(void *conn, int parc, char *parv[])
...
 
-	if (parc < 2)
+	if (parc < 4)
 	{
 		xmlrpc_generic_error(fault_needmoreparams, "Insufficient parameters.");
 		return 0;
 	}
 
-	if (*parv[0] == '#')
+	if (*parv[1] != '\0' && strlen(parv[0]) > 1)
```
aaronmdjones, line 523: The `parc` check above already guarantees that you have at least 4 parameters. This entire check is unnecessary.

modules/transport/xmlrpc/main.c
```diff
@@ -511,49 +514,68 @@ xmlrpcmethod_metadata(void *conn, int parc, char *parv[])
...
+		}
+
+		if (authcookie_validate(parv[0], mu) == false)
+		{
+			xmlrpc_generic_error(fault_badauthcookie, "Invalid authcookie for this account.");
+			return 0;
+		}
+	}
+	else
+		mu = NULL;
```
aaronmdjones, line 538: This branch can be removed due to the above.

modules/transport/xmlrpc/main.c
```diff
@@ -511,49 +514,68 @@ xmlrpcmethod_metadata(void *conn, int parc, char *parv[])
...
-		mc = mychan_find(parv[0]);
+		mc = mychan_find(parv[2]);
 		if (mc == NULL)
 		{
 			xmlrpc_generic_error(fault_nosuch_source, "No channel registration was found for the provided channel name.");
 			return 0;
 		}
 
-		md = metadata_find(mc, parv[1]);
+		md = metadata_find(mc, parv[2]);
```
aaronmdjones, line 551: This is the wrong array index.

aaronmdjones (changes requested):

modules/transport/xmlrpc/main.c
```diff
@@ -511,49 +514,63 @@ xmlrpcmethod_metadata(void *conn, int parc, char *parv[])
...
 	{
 		xmlrpc_generic_error(fault_nosuch_source, "No metadata found matching this account/channel and key.");
 		return 0;
 	}
-
-	xmlrpc_string(buf, md->value);
+	if (!strncmp(md->name, "private:", 8) && is_soper(mu))
+		xmlrpc_string(buf, md->value);
+	else
+		xmlrpc_string(buf, "");
```
aaronmdjones, line 573: Just one more thing that I missed last time.

Upon closer inspection, this has a bug; it no longer shows any metadata values unless they're private metadata and the user is a soper. Also, just checking whether the user is a soper is not granular enough; many utility bots are given soper status with only the privileges they need to perform their function (e.g. `operserv:akill`). We have specific `*:auspex` privileges for viewing private information about channels, groups, and users, so it makes sense that this should check for those privileges too.

Therefore, this block needs changing to

```C
bool showprivate = false;

if (*parv[2] == '#' && has_priv_myuser(mu, PRIV_CHAN_AUSPEX))
    showprivate = true;
else if (*parv[2] == '!' && has_priv_myuser(mu, PRIV_GROUP_AUSPEX))
    showprivate = true;
else if (*parv[2] != '#' && *parv[2] != '!' && has_priv_myuser(mu, PRIV_USER_AUSPEX))
    showprivate = true;

if (strncmp(md->name, "private:", 8) != 0 || showprivate)
    xmlrpc_string(buf, md->value);
else
    xmlrpc_string(buf, "");
```

aaronmdjones (changes requested):
Much better, with the caveat and recommended code change in my last comment.

examknow:
All should be good now. You can disregard the last commit. It was just my local version of the repo conflicting with the remote repo.

PR #756, opened 2021-02-09
implement groupserv mark

examknow:
Allow sopers to add marks to groups like they can to channels and accounts.

### DIFF
--- a/modules/groupserv/mark.c
+++ b/modules/groupserv/mark.c
@@ -1,0 +1,103 @@
+/*
+ * SPDX-License-Identifier: ISC
+ * SPDX-URL: https://spdx.org/licenses/ISC.html
+ *
+ * Copyright (C) 2005 William Pitcock
+ *
+ * Marking for channels.
+ */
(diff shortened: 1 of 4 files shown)


modules/groupserv/mark.c
```diff
@@ -0,0 +1,103 @@
+/*
+ * SPDX-License-Identifier: ISC
+ * SPDX-URL: https://spdx.org/licenses/ISC.html
+ *
+ * Copyright (C) 2005 William Pitcock
```
aaronmdjones, line 5: I highly doubt this.

modules/groupserv/mark.c
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * Marking for channels.
```
aaronmdjones, line 7: This is marking for groups.

aaronmdjones (changes requested):
Please fix the opening header on `modules/groupserv/mark.c`

PR #757, opened 2021-02-11
Improve loadable module handling

aaronmdjones:
- Remove unused functions and structure members
- Avoid dynamic memory allocation where possible
- Clarify the purpose of some variables by giving them better names
- Fix dependency tracking for `MODULE_TRY_REQUEST_DEPENDENCY()`
- Liberal const-correctness
- Replace hardcoded buffer sizes with `sizeof` and use correct data types for them
- Make following module load debug messages easier by logging that we're about to load a module
- Avoid some possible corner-case crashes
- Process `loadmodule` lines relatively
  - This is by far the largest commit in this work, and breaks backward compatibility with older configuration files.
    - Please read the message on that commit for more information.
- Support loading modules by declaring a tree of named blocks
- Unify `modules/operserv/mod{inspect,load,reload,unload}` into a new module and vastly improve those commands.
  - Please read the message on that commit for more information.


ilbelkyr (approved):
looks good to me, especially the `operserv/modulemanager` stuff I'd already looked over previously but all the other stuff too, some of which I wouldn't even have thought of.

Thank you for all the work put into this!

aaronmdjones:
All done from me.

PR #765, opened 2021-02-21
Infoserv improvements

### LINKED ISSUE #597: InfoServ: allow swapping & removing date/time from logon news
I think that on all networks that I have seen InfoServ being used on, all posts are static and sent ages ago.

Thus I would like to see InfoServ supporting not showing logon info date/time at all, but instead statically in format e.g.: 

```
-InfoServ- *** Message(s) of the Day ***
-InfoServ- [HELP] Remember that for help you can always /join #help 
-InfoServ- [DOCUMENTATION] You can find the documentation at https://wiki.example.net/
-InfoServ- [FIREWALL] Please run software firewall on your host while using IRC
-InfoServ- *** End of Message(s) of the Day ***
```

Where there is just the MOTD begin, subjects and the messages rather than

```
-InfoServ- *** Message(s) of the Day ***
-InfoServ- [SUBJECT] Notice from Mikaela, posted 08:08 on 11/21/2017:
-InfoServ- BLAH BLAH BLAH
-InfoServ- [SUBJECT2] Notice from Mikaela, posted 08:08 on 11/21/2017:
-InfoServ- BLAH BLAH BLAH BLAH
-InfoServ- [SUBJECT3] Notice from Mikaela, posted 08:08 on 11/21/2017:
-InfoServ- BLAH BLAH BLAH BLAH BLAH
-InfoServ- *** End of Message(s) of the Day ***
```

I think that simultaneously it would be useful to be able to manage the order in which InfoServ logon news are sent without having to remove all posts and writing them again.

ilbelkyr:
Fixes #597, which inspired this PR to allow reordering entries and add an option to display just the entries without their setter and timestamp information. It ended up including a few other improvements as well.

### DIFF
--- a/modules/infoserv/main.c
+++ b/modules/infoserv/main.c
@@ -15,20 +15,15 @@
 
 #include <atheme.h>
 
+#define DB_TYPE_LOGONINFO "LI"
+#define DB_TYPE_LOGONINFO_OPER "LIO"
+
 struct logoninfo
 {
-        stringref nick;
-        char *subject;
-        time_t info_ts;
-        char *story;
-};
-
-struct operlogoninfo
-{
-        stringref nick;
-        char *subject;
-        time_t info_ts;
-        char *story;
+	stringref nick;
+	char *subject;
+	time_t info_ts;
+	char *story;
 };
 
 static mowgli_list_t logon_info;
@@ -66,32 +61,29 @@ underscores_to_spaces(char *y)
 }
 
 static void
+write_infoentry(struct database_handle *db, struct logoninfo *l, bool oper)
+{
+	db_start_row(db, oper ? DB_TYPE_LOGONINFO_OPER : DB_TYPE_LOGONINFO);
+	db_write_word(db, l->nick);
+	db_write_word(db, l->subject);
+	db_write_time(db, l->info_ts);
+	db_write_str(db, l->story);
+	db_commit_row(db);
+}
+
+static void
 write_infodb(struct database_handle *db)
 {
 	mowgli_node_t *n;
 
 	MOWGLI_ITER_FOREACH(n, logon_info.head)
 	{
-		struct logoninfo *l = n->data;
-
-		db_start_row(db, "LI");
-		db_write_word(db, l->nick);
-		db_write_word(db, l->subject);
-		db_write_time(db, l->info_ts);
-		db_write_str(db, l->story);
-		db_commit_row(db);
+		write_infoentry(db, n->data, false);
 	}
 
 	MOWGLI_ITER_FOREACH(n, operlogon_info.head)
 	{
-		struct operlogoninfo *o = n->data;
-
-		db_start_row(db, "LIO");
-		db_write_word(db, o->nick);
-		db_write_word(db, o->subject);
-		db_write_time(db, o->info_ts);
-		db_write_str(db, o->story);
-		db_commit_row(db);
+		write_infoentry(db, n->data, true);
 	}
 
 }
@@ -109,38 +101,17 @@ db_h_li(struct database_handle *db, const char *type)
 	l->subject = sstrdup(subject);
 	l->info_ts = info_ts;
 	l->story = sstrdup(story);
-	mowgli_node_add(l, mowgli_node_create(), &logon_info);
+
+	if (strcasecmp(type, DB_TYPE_LOGONINFO_OPER) == 0)
+		mowgli_node_add(l, mowgli_node_create(), &operlogon_info);
+	else
+		mowgli_node_add(l, mowgli_node_create(), &logon_info);
 }
 
 static void
-db_h_lio(struct database_handle *db, const char *type)
+display_info(struct user *u, bool oper)
 {
-	const char *nick = db_sread_word(db);
-	const char *subject = db_sread_word(db);
-	time_t info_ts = db_sread_time(db);
-	const char *story = db_sread_str(db);
-
-	struct operlogoninfo *const o = smalloc(sizeof *o);
-	o->nick = strshare_get(nick);
-	o->subject = sstrdup(subject);
-	o->info_ts = info_ts;
-	o->story = sstrdup(story);
-	mowgli_node_add(o, mowgli_node_create(), &operlogon_info);
-}
-
-static void
-display_info(struct hook_user_nick *data)
-{
-	struct user *u;
-	mowgli_node_t *n;
-	struct logoninfo *l;
-	char dBuf[BUFSIZE];
-	struct tm *tm;
-	unsigned int count = 0;
-
-	u = data->u;
-	if (u == NULL)
-		return;
+	return_if_fail(u != NULL);
 
 	// abort if it's an internal client
 	if (is_internal_client(u))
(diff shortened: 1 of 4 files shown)


modules/infoserv/main.c
```diff
@@ -287,35 +258,27 @@ is_cmd_post(struct sourceinfo *si, int parc, char *parv[])
...
-		n = mowgli_node_create();
-		mowgli_node_add(l, n, &logon_info);
-	}
+	mowgli_list_t *list = (imp == 0) ? &operlogon_info : &logon_info;
+	mowgli_node_t *n = mowgli_node_create();
+	mowgli_node_add(l, n, list);
 
 	command_success_nodata(si, _("Added entry to logon info"));
+	if (logoninfo_count != 0 && !logoninfo_reverse && list->count > logoninfo_count)
+		command_success_nodata(si, ngettext("Warning: There is more than %u entry, so your message will not be displayed by default.",
```
aaronmdjones, line 273: These strings need to be wrapped in `N_()` so that the `update-pot` target in `po/` can find them.

modules/infoserv/main.c
```diff
@@ -326,159 +289,250 @@ is_cmd_post(struct sourceinfo *si, int parc, char *parv[])
...
 	}
 
+	unsigned int id;
 	if (! string_to_uint(target, &id) || ! id)
 	{
-		command_fail(si, fault_badparams, STR_INVALID_PARAMS, "DEL");
-		command_fail(si, fault_badparams, _("Syntax: DEL <id>"));
+		command_fail(si, fault_badparams, STR_INVALID_PARAMS, oper ? "ODEL" : "DEL");
+		if (oper)
+			command_fail(si, fault_needmoreparams, _("Syntax: ODEL <id>"));
```
aaronmdjones, line 311: Wrong fault code (copy/paste mistake).

modules/infoserv/main.c
```diff
@@ -326,159 +289,250 @@ is_cmd_post(struct sourceinfo *si, int parc, char *parv[])
...
+		else
+			command_fail(si, fault_needmoreparams, _("Syntax: DEL <id>"));
```
aaronmdjones, line 313: Likewise.

modules/infoserv/main.c
```diff
@@ -326,159 +289,250 @@ is_cmd_post(struct sourceinfo *si, int parc, char *parv[])
...
-			return;
+	unsigned int pos = 0;
+	MOWGLI_ITER_FOREACH(n, list->head)
+	{
+		// 1-based indexing, reflecting what we display in LIST output
+		pos++;
+
+		if (!from_node && pos == from_id)
+		{
+			// We'll be removing this node from the list and adding it before the
```
aaronmdjones, line 426: Please use `/* */` for multiline comments.

modules/infoserv/main.c
```diff
@@ -326,159 +289,250 @@ is_cmd_post(struct sourceinfo *si, int parc, char *parv[])
...
 		}
+
+		if (!to_node && pos == to_id)
+			to_node = n;
+
+		if (from_node && to_node)
+			break;
+	}
+
+	// this should never happen
```
aaronmdjones, line 442: Likewise.

modules/infoserv/main.c
```diff
@@ -326,159 +289,250 @@ is_cmd_post(struct sourceinfo *si, int parc, char *parv[])
...
 
-	command_fail(si, fault_nosuch_target, _("Entry \2%u\2 not found in oper logon info."), id);
+	mowgli_node_delete(from_node, list);
+	mowgli_node_add_before(entry, from_node, list, to_node);
+
+	if (oper)
+		command_success_nodata(si, _("Oper logon info entry \2%u\2 moved to position \2%u\2."), from_id, to_id);
+	else
+		command_success_nodata(si, _("Logon info entry \2%u\2 moved to position \2%u\2."), from_id, to_id);
+	logcommand(si, CMDLOG_ADMIN, "INFO:%s: \2%u\2 -> \2%u\2: \2%s\2, \2%s\2", oper ? "OMOVE" : "MOVE", from_id, to_id, entry->subject, entry->story);
```
aaronmdjones, line 463: Is it worth logging the notice contents (`entry->story`) here?

And how about we log the subject first, a la
```C
logcommand(si, CMDLOG_ADMIN, "INFO:%s: \2%s\2: \2%u\2 -> \2%u\2", oper ? "OMOVE" : "MOVE", entry->subject, from_id, to_id);
```

modules/infoserv/main.c
```diff
@@ -546,21 +618,46 @@ mod_init(struct module *const restrict m)
...
 
+	struct infoserv_persist_record *rec = mowgli_global_storage_get(INFOSERV_PERSIST_STORAGE_NAME);
+
+	if (rec)
+	{
+		mowgli_global_storage_free(INFOSERV_PERSIST_STORAGE_NAME);
+
+		if (rec->version > INFOSERV_PERSIST_VERSION)
+		{
+			slog(LG_ERROR, "infoserv/main: attempted downgrade is not supported (from %d to %d)", rec->version, INFOSERV_PERSIST_VERSION);
```
aaronmdjones, line 629: Use `m->name` here.

modules/infoserv/main.c
```diff
@@ -546,21 +618,46 @@ mod_init(struct module *const restrict m)
...
+			m->mflags = MODFLAG_FAIL;
```
aaronmdjones, line 630: This should be a bitwise OR.

modules/infoserv/main.c
```diff
@@ -326,159 +289,250 @@ is_cmd_post(struct sourceinfo *si, int parc, char *parv[])
...
+
+static void
+is_list_impl(struct sourceinfo *si, int parc, char *parv[], bool oper)
 {
-	struct logoninfo *l;
 	mowgli_node_t *n;
-	struct tm *tm;
-	char dBuf[BUFSIZE];
 	unsigned int x = 0;
+	mowgli_list_t list = oper ? operlogon_info : logon_info;
```
aaronmdjones, line 484: This makes the compiler emit a pointless memory copy for struct assignment; take its address and use a pointer instead.

modules/infoserv/main.c
```diff
@@ -570,7 +667,35 @@ mod_init(struct module *const restrict m)
...
+	db_unregister_type_handler(DB_TYPE_LOGONINFO_OPER);
+
+	service_unbind_command(infoserv, &is_help);
+	service_unbind_command(infoserv, &is_post);
+	service_unbind_command(infoserv, &is_del);
+	service_unbind_command(infoserv, &is_odel);
+	service_unbind_command(infoserv, &is_move);
+	service_unbind_command(infoserv, &is_omove);
+	service_unbind_command(infoserv, &is_list);
+	service_unbind_command(infoserv, &is_olist);
```
aaronmdjones, line 696: These are unnecessary because `service_delete()` now [clears out the command trie](https://github.com/atheme/atheme/blob/master/libathemecore/servtree.c#L382-L383) itself (commit 2a9d68b63285ce77d157).

aaronmdjones (changes requested):

aaronmdjones:
The rest looks okay to me, but it does (still) have some pointless extra dynamic memory allocation.

A `struct logoninfo` can only ever be in one `mowgli_list_t`, so it may be worth adding an intrusive `mowgli_node_t` in the former and removing the corresponding `mowgli_node_create()` and `mowgli_node_free()` calls.

modules/infoserv/main.c
ilbelkyr: `xgettext` handles `ngettext()` calls by default, including proper handling of the pluralization stuff. That said, it does look like our `update-potfiles.sh` script doesn't check for calls to it. In practice, most files using `ngettext()` will also use `_()` somewhere in the code, but we probably should add a check for that.

modules/infoserv/main.c
ilbelkyr: I've mostly tried to keep this analogous to the existing log messages for `DEL`/`ODEL`, keeping in mind that the subject alone is not enough to identify an entry. (In fact, the only reliable way to identify an entry in the presence of complete duplicates would be its position, but that isn't very suitable as an identifier given it shifts around when entries are moved/deleted.)

Generally, though, I'd agree that it would be ideal if we had some sort of identifier that we could log first (before the `from -> to` part of the log message).

(Whether it'd be a good idea or not, requiring the subject to be unique (and thus allowing it to serve as an identifier) would break compatibility with existing setups that have multiple entries with the same subject (due to using it as more of a "category" field, perhaps).)

modules/infoserv/main.c
aaronmdjones: My bad; I misunderstood the documentation for the `--keyword` option in `xgettext(1)`. If you specify it with no value, *then* it inhibits the default keywords like `ngettext`. However, we only specify *additional* keywords, so it will (and does) find these strings.

modules/infoserv/main.c
aaronmdjones: Fair enough.

aaronmdjones (approved):

PR #767, opened 2021-02-28
Support rehashing of `general::commit_interval`

aaronmdjones:


modules/operserv/readonly.c
```diff
@@ -55,6 +62,12 @@ os_cmd_readonly(struct sourceinfo *si, int parc, char *parv[])
 		wallops("\2%s\2 unset the READONLY option.", get_oper_name(si));
 		logcommand(si, CMDLOG_ADMIN, "READONLY:OFF");
 		command_success_nodata(si, _("Read-only mode is now disabled."));
+
+		if (commit_interval_timer)
+			(void) mowgli_timer_destroy(base_eventloop, commit_interval_timer);
+
+		commit_interval_timer = mowgli_timer_add(base_eventloop, "db_save_periodic", &db_save_periodic, NULL,
+		                                         config_options.commit_interval);
```
ilbelkyr, line 70: This doesn't check if `db_save` is set, thus scheduling a null pointer dereference if running without a write-capable backend.

Arguably, the `READONLY` command ought to fail early in that case (similar to the check added in `operserv/set_commitinterval`) – there's little point toggling database writes when they are impossible anyway.

ilbelkyr (changes requested):
Other than the inline comment above, this looks good and happens to fix a bug where the `operserv/readonly` module didn't actually appear to stop periodic database writes at all.

modules/operserv/set_commitinterval.c
```diff
@@ -22,20 +22,43 @@ os_cmd_set_commitinterval_func(struct sourceinfo *const restrict si, const int p
 		return;
 	}
 
+	if (! db_save)
+	{
+		(void) command_fail(si, fault_badparams, _("You have no write-capable database backend loaded."));
```
ilbelkyr, line 27: `fault_badparams` doesn't seem right here, though I'm not sure what the best choice would be tbh

modules/operserv/readonly.c
```diff
@@ -22,6 +22,12 @@ os_cmd_readonly(struct sourceinfo *si, int parc, char *parv[])
 		return;
 	}
 
+	if (! db_save)
+	{
+		(void) command_fail(si, fault_badparams, _("You have no write-capable database backend loaded."));
```
ilbelkyr, line 27: (see other `fault_badparams` comment)

ilbelkyr (changes requested):
This should be fine, but we'll need to bikeshed the appropriate faultcode used if there's no suitable `db_write`

aaronmdjones:
Perhaps `fault_unimplemented`?

ilbelkyr:
Either that or `fault_internalerror`, although that one feels like more of a "things went wrong and it's our/the network's fault" kind of thing.

I suppose `fault_unimplemented` was originally meant for things that are supposed to work but aren't implemented, indicating an issue with the implementation – we barely make use of it as it is, though; the current uses of it include:

- `gameserv/dice` running across a math operator it doesn't know. (Feels like *that* should be `fault_badparams` instead, tbh)
- `chanserv/access` trying to modify founder status of a role. That one might be arguable?
- `contrib/cs_access_alias` failing to find the `FLAGS` command to fall back to. That one probably should have a proper dependency, and `fault_internalerror` if it somehow fails anyway.
- `scripting/perl` on an "unexpected error". I feel that ought to be `fault_internalerror`.
- `scripting/perl` failing to provide help because "Perl help commands [are] not yet implemented". Okay, that's the poster child, I guess.
- `transport/jsonrpc`, `transport/xmlrpc` when a command doesn't send a reply via the proper functions. Not sure it's a good fit here, either.

Granted, this isn't really the right issue to take apart questionably defined fault codes and we should look into that another time, perhaps.

ilbelkyr:
@aaronmdjones let's go with `fault_unimplemented` for now and we can sort through the other thoughts I mentioned at a separate point, maybe?

ilbelkyr (approved):

PR #771, opened 2021-03-28
Add Graphite Exporter module

aaronmdjones:
This module collects and exports statistics to a [Graphite statistics server](https://graphiteapp.org/).

At the moment, it can be used for graphing:

- Number of registered accounts, channels, and groups in the services database
- Number of users, channels, and servers on the network
- Number of types of password hash (useful for keeping track of a crypto module migration)


aaronmdjones:
The password hash functionality could be made much more efficient if there were a `user_changed_password` hook that it could plug into, so that it could cache what type of password hash each user has and invalidate that cache properly.

Do you think it's worth it?

alyx:
> The password hash functionality could be made much more efficient if there were a `user_changed_password` hook that it could plug into, so that it could cache what type of password hash each user has and invalidate that cache properly.
> 
> Do you think it's worth it?

IMO More hooks is better, opens up more doors for other things as well in the future and makes this more efficient

aaronmdjones:
It has caching now.

PR #782, opened 2021-05-29
libathemecore/ptasks.c: STATS:T: make nonsensitive data available

aaronmdjones:
libathemecore/ptasks.c: STATS:T: make nonsensitive data available

The statistics for `server`, `user`, and `chan` can already be obtained via the `LUSERS` command on IRC, so it does not make sense for services to prevent accessing these by requiring services privileges.

Furthermore, the number of registered accounts, nicknames, and channels should also not be considered sensitive data. The Atheme project does not condone hiding such basic and non-sensitive data from network users.

Finally, put missing `mynicks` count in default MOTD file.

### DIFF
--- a/dist/atheme.motd.example
+++ b/dist/atheme.motd.example
@@ -4,6 +4,7 @@ Nickname registrations will expire after &nickexpiry& days.
 Channel registrations will expire after &chanexpiry& days.
 
 Registered users: &myusers&
+Registered nicks: &mynicks&
 Registered channels: &mychans&
 
 The services administrator can add additional information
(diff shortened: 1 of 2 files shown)


Xe (approved):

alyx (approved):

aaronmdjones:
It comes to mind that there is another possibility; to keep only one block for both `t` and `T` (as before this change), but re-arrange the non-sensitive data points to be emitted before the permission check.

Thoughts? I'm not sure which approach I like more. It's not like there's no precedent for treating case differently, though; `u` has no `U` sibling.

aaronmdjones:
I decided to go with that. Initial comment updated.

kaniini (approved):

mweinelt (approved):

ilbelkyr (approved):

alyx (approved):

PR #783, opened 2021-05-29
modules/protocol: delete elemental-ircd.c

Xe:
This IRCd is abandoned: https://christine.website/blog/deprecation-notice-elemental-ircd-2019-02-11


aaronmdjones:
You forgot to edit `modules/protocol/Makefile`

alyx:
@Xe what if we _don't_ remove it and you just rebase elemental off of solanum and call it a live ircd instead? 👉 👈 🥺

alyx:
Also on a serious note, @aaronmdjones I think I've brought this up before, but thoughts on instead of deleting protocols for older/abandoned IRCds, moving them out of core to like "contrib/legacy/..." or something? I'd brought this up a few years back, but we've been rather lacking in dead IRCds. Generally it's rare for dead IRCds to cause much extra development work for us, aside from updating headers and such if we change them, and having it available would mean if someone specifically wanted to mess around with legacy platforms, they wouldn't have to dig through "Okay which Atheme release was this" to do so.

Xe:
At the very least if it's kept around I'd like my name and email to be updater.

PR #790, opened 2021-06-19
Make VERIFY EMAILCHG confirm the account if necessary

edk0:
Currently you can verify an email address change on an unconfirmed account, but it won't confirm the account even though you've just proved that you own its current email address. Everyone I've spoken to about this thinks it's a bug, so here's a fix.

I've also changed the log message to identify the target of the verify, rather than repeating the name of the requestor; morally that should be its own PR, but it would have had a conflict with this one.

### DIFF
--- a/modules/nickserv/verify.c
+++ b/modules/nickserv/verify.c
@@ -66,12 +66,12 @@ ns_cmd_verify(struct sourceinfo *si, int parc, char *parv[])
 
 		if (!strcasecmp(key, md->value))
 		{
-			logcommand(si, CMDLOG_SET, "VERIFY:REGISTER: \2%s\2 (email: \2%s\2)", get_source_name(si), mu->email);
+			logcommand(si, CMDLOG_SET, "VERIFY:REGISTER: \2%s\2 (email: \2%s\2)", entity(mu)->name, mu->email);
 			ns_verify_activate_account(si, mu);
 			return;
 		}
 
-		logcommand(si, CMDLOG_SET, "failed VERIFY REGISTER \2%s\2, \2%s\2 (invalid key)", get_source_name(si), mu->email);
+		logcommand(si, CMDLOG_SET, "failed VERIFY REGISTER \2%s\2, \2%s\2 (invalid key)", entity(mu)->name, mu->email);
 		command_fail(si, fault_badparams, _("Verification failed. Invalid key for \2%s\2."),
 			entity(mu)->name);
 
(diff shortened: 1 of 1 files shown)


aaronmdjones:
I've already separately fixed this a few weeks ago, including your log message change and making it a cmdtree so that other modules can add more `VERIFY` subcommands. It's awaiting a review from @ilbelkyr

edk0:
oops

modules/nickserv/verify.c
```diff
@@ -104,18 +108,23 @@ ns_cmd_verify(struct sourceinfo *si, int parc, char *parv[])
...
 			myuser_set_email(mu, md->value);
 
-			logcommand(si, CMDLOG_SET, "VERIFY:EMAILCHG: \2%s\2 (email: \2%s\2)", get_source_name(si), mu->email);
+			logcommand(si, CMDLOG_SET, "VERIFY:EMAILCHG: \2%s\2 (email: \2%s\2)", entity(mu)->name, mu->email);
 
 			metadata_delete(mu, "private:verify:emailchg:key");
 			metadata_delete(mu, "private:verify:emailchg:newemail");
 			metadata_delete(mu, "private:verify:emailchg:timestamp");
 
 			command_success_nodata(si, _("\2%s\2 has now been verified."), mu->email);
```
aaronmdjones, line 117: Please remove this line.

aaronmdjones (changes requested):
No worries. I can merge this right now and still carry on with my new version in a bit when they return. It was she who wanted the cmdtree thing anyway. First, though, I'll need you to make the following change, otherwise it will produce duplicate output.

modules/nickserv/verify.c
```diff
@@ -202,6 +192,11 @@ ns_cmd_fverify(struct sourceinfo *si, int parc, char *parv[])
 
 		command_success_nodata(si, _("\2%s\2 has now been verified."), mu->email);
```
aaronmdjones, line 193: And this one.

edk0:
I left those in on purpose since two lines reflect different things -- one's about the email address and the other is about the account. But if we don't want that I think they'd need to be wrapped in an else rather than just going away completely?

aaronmdjones:
Hmm. In my version, I just log the email address, e.g. an unconditional:

```C
command_success_nodata(si, _("\2%s\2 has now been verified."), mu->email);
```

... because it is the e-mail address you're verifying control over either way.

edk0:
Hmm. I think I'd like to be explicit that they're verifying both things (and it would make sense that everyone gets to see the thank you message once, too), but maybe we could roll them up into a single message?

aaronmdjones:
I will merge this as-is and rebase my work on it. I'll also make mine state the account name and e-mail in the same message.

PR #791, opened 2021-06-20
unreal4 wants SID, not SERVER, for remote server introductions

jesopo:
draft because this has only been tested live on pissnet so I'd like to run it by people more responsible than me


modules/protocol/unreal4.c
```diff
@@ -617,12 +617,41 @@ static void
...
+					return;
+
+				continue;
+			}
+			else if (sid[i] == '9')
+				sid[i] = 'A';
+			else sid[i]++;
+			break;
+		}
+	} while (server_find(sid));
```
kaniini, line 652: generally we like to be explicit about the NULL check

kaniini (approved):
this is correct other than the style change (or at least it was back in the ustream days)

modules/protocol/unreal4.c
tzvetkoff: I do happen to see a few potential problems here:

1. This algorithm will try generating `sid`s like `A00` after `9ZZ`.
2. `sid` is static, meaning it will eventually exceed `9ZZ` (even though nobody actually will run services _that_ long and jupe that many servers.)

I also wonder why don't reuse the original server `sid` (_if found_) and only generate a new `sid` when one is proactively jupe'ing servers before they were even introduced?

jesopo: this SID generation was directly stolen from the other protocol modules

jesopo:
I think this is currently in a state that is technically correct and works but it will still fail on atypically complex networks (i.e. pissnet) and I'm not certain why. I think we ought to be doing this via `SF_JUPE_PENDING` a-la bahamut & inspircd but I've not been able to make that work; every time I defer an unreal4 jupe with this, atheme suddenly stops processing (or even debug logging) received `SQUIT`s

PR #798, opened 2021-06-27
Fix password overflow error message to mention bytes instead of characters

progval:
For example, when using this 150-characters password:

```
C -> S: PRIVMSG NickServ :REGISTER éééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééééé foo@example.org
S -> C: :nickserv!nickserv@services.int NOTICE foo :Invalid parameters for REGISTER.
S -> C: :nickserv!nickserv@services.int NOTICE foo :Registration passwords may not be longer than 288 characters.
```

(Same bug on Anope: https://github.com/anope/anope/pull/275 ^^)

### DIFF
--- a/modules/nickserv/register.c
+++ b/modules/nickserv/register.c
@@ -60,7 +60,7 @@ ns_cmd_register(struct sourceinfo *si, int parc, char *parv[])
 	if (strlen(pass) > PASSLEN)
 	{
 		command_fail(si, fault_badparams, STR_INVALID_PARAMS, "REGISTER");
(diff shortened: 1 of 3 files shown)


alyx:
I get what you mean here but I don't really love the phrasing "bytes" -- for a non-technical user, that's not particularly clear. Not precisely sure what would be a better phrasing, offhand, but I think it might be worth playing around with.

ProgVal:
I agree it's not great, but I don't see a better way to phrase it. The implementation is intrinsically hard to explain

PR #806, opened 2021-07-02
support HS TAKE without params if there's only one HS OFFER

### LINKED ISSUE #804: HS TAKE without params should succeed if there's only 1 OFFER

jesopo:
closes #804

might not be the best way to do this, and the help text will still show `<vhost>` rather than `[vhost]`. thoughts?

### DIFF
--- a/modules/hostserv/offer.c
+++ b/modules/hostserv/offer.c
@@ -262,7 +262,7 @@ hs_cmd_take(struct sourceinfo *si, int parc, char *parv[])
 	struct metadata *md;
 	time_t vhost_time = 0;
 
-	if (!host)
+	if (!host && MOWGLI_LIST_LENGTH(&hs_offeredlist) > 1)
 	{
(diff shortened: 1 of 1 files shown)


modules/hostserv/offer.c
```diff
@@ -298,7 +298,7 @@ hs_cmd_take(struct sourceinfo *si, int parc, char *parv[])
 		if (l->group != NULL && !myuser_is_in_group(si->smu, l->group))
 			continue;
 
-		if (!irccasecmp(l->vhost, host))
+		if (!host || !irccasecmp(l->vhost, host))
```
spb, line 301: This will behave unexpectedly if there's one vhost offered, but it's restricted by group. I think a proper solution needs a bit more rearranging.

spb: I've created #807 with what I think is a more complete way to do this.

PR #807, opened 2021-07-02
Allow HS TAKE without params if there's only one HS OFFER (#2)

spb:
I think this is a more complete approach to #804, and handles the case where there's a single offer but it's restricted by group. Not properly tested because I don't have a local ircd any more.

### DIFF
--- a/modules/hostserv/offer.c
+++ b/modules/hostserv/offer.c
@@ -252,6 +252,22 @@ myuser_is_in_group(struct myuser *mu, struct myentity *mt)
 	return false;
 }
 
+static void
+take_vhost(struct sourceinfo *si, const char *host)
+{
+	if (strstr(host, "$account"))
+		replace(host, BUFSIZE, "$account", entity(si->smu)->name);
+
+	if (!check_vhost_validity(si, host))
+		return;
+
+	logcommand(si, CMDLOG_GET, "TAKE: \2%s\2 for \2%s\2", host, entity(si->smu)->name);
+
+	command_success_nodata(si, _("You have taken vhost \2%s\2."), host);
+	hs_sethost_all(si->smu, host, get_source_name(si));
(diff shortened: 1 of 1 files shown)


modules/hostserv/offer.c
```diff
@@ -255,8 +255,13 @@ myuser_is_in_group(struct myuser *mu, struct myentity *mt)
...
 take_vhost(struct sourceinfo *si, const char *host)
 {
+	char local_host[HOSTLEN+1];
 	if (strstr(host, "$account"))
-		replace(host, BUFSIZE, "$account", entity(si->smu)->name);
+	{
+		strncpy(local_host, host, HOSTLEN);
+		replace(local_host, BUFSIZE, "$account", entity(si->smu)->name);
+		host = local_host;
+	}
```
aaronmdjones: This should use `mowgli_strlcpy()` instead of `strncpy(3)`, and the buffer should be `BUFSIZE` bytes large. At the moment it may crash because you're telling `replace()` that it's `BUFSIZE` bytes large, when it isn't. The `check_vhost_validity()` function will already do a length test.

aaronmdjones (changes requested):

modules/hostserv/offer.c
```diff
@@ -252,6 +252,27 @@ myuser_is_in_group(struct myuser *mu, struct myentity *mt)
 	return false;
 }
 
+static void
+take_vhost(struct sourceinfo *si, const char *host)
+{
+	char local_host[BUFSIZE];
+	if (strstr(host, "$account"))
+	{
+		mowgli_strlcpy(local_host, host, HOSTLEN);
```
aaronmdjones, line 261: Not quite; `BUFSIZE`. The third parameter is the size of the destination; it will already terminate early upon copying a null character.

aaronmdjones (approved):

PR #811, opened 2021-07-08
Update chanserv akick help to be clearer

tsposato:
The !P isn't required for an akick ban to be permanent, so I've removed it from the help.
Not sure if someone wants to also remove it from the codebase as well if deemed necessary.
I think this makes it clearer and easier to add a permanent vs. timed akick.

### DIFF
--- a/help/default/cservice/akick
+++ b/help/default/cservice/akick
@@ -6,7 +6,7 @@ automatically kickbanned when they join the channel,
 removing any matching ban exceptions first. Users
 with the +r flag are exempt.
 
-Syntax: AKICK <#channel> ADD <nickname|hostmask> [!P|!T <minutes>] [reason]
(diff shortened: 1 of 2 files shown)


ilbelkyr:
Without either `!T <duration>` or `!P`, akicks are set to the default expiry – this defaults to permanent, but is configurable. See the relevant section of the configuration file,

https://github.com/atheme/atheme/blob/dcc403ae23466231b8c6a4b590b0fc7bb02f76be/dist/atheme.conf.example#L1936-L1941

(We shouldn't remove the `!P` syntax itself either way IMO – even if it is a no-op with a particular configuration, supporting it when specifying a duration avoids breaking compatibility plus allows people to avoid relying on the configured default)

Adding a mention of the default to the help seems like a good idea, though would require touching the help code to be able to list the configured value. (This would ideally be done in a more general way that's not specific to this config option alone.) I'd recommend opening a separate issue about this where we can consider the best approach to doing so.

PR #816, opened 2021-08-14
don't just ignore invalid CS LIST criteria/parameters

### LINKED ISSUE #801: In CS LIST, return an error if a criteron is invalid instead of ignoring it and continuing
When parsing the criteria passed to CS LIST, Atheme removes all invalid criteria and then runs the query with the remaining criteria. For example:

`LIST MLOCK +i INVALIDCRITERION`

becomes:

`LIST MLOCK +i`

and:

`LIST #example*`

becomes:

`LIST`

This gets extremely cumbersome when there are several thousand channels registered and sopers forget to do `LIST PATTERN foo` instead of `LIST foo`, either because they didn't drink coffee yet or because they're used to Atheme syntax that went away 11 years ago. There is (thankfully) an output limit, but I suggest Atheme should just return an error (e.g. "Invalid criterion #example*" for the latter example) instead of removing invalid criteria and continuing.

jesopo:
I'm certain there's going to be something about how I chose to pass the information back from `process_parvarray()` that isn't right, but it works!

closes #801

### DIFF
--- a/modules/chanserv/list.c
+++ b/modules/chanserv/list.c
@@ -21,6 +21,7 @@ enum list_opttype
 
 enum list_opterr
 {
+	OPTERR_NONE,
 	OPTERR_UNKNOWN_OPT,
 	OPTERR_BAD_ARG,
 };
@@ -119,7 +120,7 @@ process_parvarray(const struct list_option *opts, size_t optsize, int parc, char
 		if (!found)
 			return OPTERR_UNKNOWN_OPT;
 	}
-	return 0;
+	return OPTERR_NONE;
 }
 
 static void
--- a/modules/chanserv/list.c
+++ b/modules/chanserv/list.c
@@ -56,18 +56,21 @@ parse_age(const char *s)
 	return duration;
 }
 
-static void
-process_parvarray(const struct list_option *opts, size_t optsize, int parc, char *parv[])
+static unsigned short
+process_parvarray(const struct list_option *opts, size_t optsize, int parc, char *parv[], char **opt_last)
 {
 	int i;
 	size_t j;
+	bool found = 0;
 
 	for (i = 0; i < parc; i++)
 	{
+		*opt_last = parv[i];
 		for (j = 0; j < optsize; j++)
 		{
 			if (!strcasecmp(opts[j].option, parv[i]))
 			{
+				found = 1;
 				switch(opts[j].opttype)
 				{
 				case OPT_BOOL:
@@ -79,6 +82,8 @@ process_parvarray(const struct list_option *opts, size_t optsize, int parc, char
 						*opts[j].optval.intval = atoi(parv[i + 1]);
 						i++;
 					}
+					else
+						return 2;
 					break;
 				case OPT_STRING:
 					if (i + 1 < parc)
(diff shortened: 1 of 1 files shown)


jesopo:
```
16:59 <jess> list a
16:59 -ChanServ(ChanServ@services.int)- Error: a is not a valid LIST option
16:59 <jess> list pattern
16:59 -ChanServ(ChanServ@services.int)- Error: Invalid argument for option pattern
16:59 <jess> list pattern * a
16:59 -ChanServ(ChanServ@services.int)- Error: a is not a valid LIST option
16:59 <jess> list pattern *
16:59 -ChanServ(ChanServ@services.int)- Channels matching pattern * :
16:59 -ChanServ(ChanServ@services.int)- No channel matched criteria pattern *
```

glguy:
You'll need to add your success enum case to your result type, currently `0` has two meanings, it's either the first error or success.

```
➜  ~ cat test.c
#include <stdio.h>
enum list_opterr { OPTERR_UNKNOWN_OPT, OPTERR_BAD_ARG, };
int main(void) {
        printf("OPTERR_UNKNOWN_OPT (%d)\n", OPTERR_UNKNOWN_OPT);
        printf("OPTERR_BAD_ARG (%d)\n", OPTERR_BAD_ARG);
}
➜  ~ clang test.c -o test && ./test
OPTERR_UNKNOWN_OPT (0)
OPTERR_BAD_ARG (1)
```

something like

```c
enum list_result
{
	OPTSUCCESS,
	OPTERR_UNKNOWN_OPT,
	OPTERR_BAD_ARG,
};

static enum list_result
process_parvarray(const struct list_option *opts, size_t optsize, int parc, char *parv[], char **opt_last)
```

jesopo:
@glguy looking better?

modules/chanserv/list.c
```diff
@@ -56,18 +63,22 @@ parse_age(const char *s)
 	return duration;
 }
 
-static void
-process_parvarray(const struct list_option *opts, size_t optsize, int parc, char *parv[])
+static unsigned int
```
dwfreed, line 66:
```suggestion
static enum list_opterr
```

Since you only return enum values, the return type should be the enum.

modules/chanserv/list.c
```diff
@@ -190,7 +210,16 @@ cs_cmd_list(struct sourceinfo *si, int parc, char *parv[])
 	if (si->c != NULL)
 		return;
 
-	process_parvarray(optstable, ARRAY_SIZE(optstable), parc, parv);
+	char *opt_last;
+	unsigned int parv_err = process_parvarray(optstable, ARRAY_SIZE(optstable), parc, parv, &opt_last);
```
dwfreed, line 214:
```suggestion
	enum list_opterr parv_err = process_parvarray(optstable, ARRAY_SIZE(optstable), parc, parv, &opt_last);
```

Likewise

dwfreed (changes requested):

aaronmdjones (approved):

aaronmdjones:
`opt_last` is uninitialised if the function returns no error, but the caller does not use its value in that case anyway, so that's okay.

PR #817, opened 2021-08-16
canonicalise emails before checking them against BADMAIL

### LINKED ISSUE #773: canonicalise badmail

jesopo:
closes #773

### DIFF
--- a/modules/nickserv/badmail.c
+++ b/modules/nickserv/badmail.c
@@ -62,11 +62,13 @@ check_registration(struct hook_user_register_check *hdata)
 	if (hdata->approved)
 		return;
 
+	stringref email_canonical = canonicalize_email(hdata->email);
+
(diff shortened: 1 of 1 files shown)


jesopo:
maybe worth checking both noncanonicalised and canonicalised?

jesopo:
`<@amdj> jess: #817 leaks memory; you need to strshare_unref() the result before returning.`

spb:
> maybe worth checking both noncanonicalised and canonicalised?

Better to canonicalise the badmail entries, IMO.

jesopo:
no, we get temporary mail providers that use dots in gmail addreses to differentiate addresses and I'd like to badmail `*.*.*.*.*@gmail.com`

spb:
OK, fair point. In that case you'll have to test both versions.

PR #818, opened 2021-08-16
only call ircd_on_logout when the ircd knows a user is logged in

### LINKED ISSUE #708: freenode user generated `:nick ACCOUNT *` from dropping an unverified account
as freenode/ircd-seven/charybdis uses a hack of unverified users not actually having an announced account to make them not qualify for registered-users-only mechanisms, it makes an IRCv3 `account-notify` logout message not make sense as no one yet thought they had an account.

jesopo:
closes #708. annoying that I had to do it in so many places but I didn't see another option

### DIFF
--- a/libathemecore/services.c
+++ b/libathemecore/services.c
@@ -413,7 +413,8 @@ bool ircd_logout_or_kill(struct user *u, const char *login)
 	hook_call_user_can_logout(&req);
 
 	if (req.allowed)
-		return ircd_on_logout(u, login);
+		if (!(u->myuser->flags & MU_WAITAUTH))
+			return ircd_on_logout(u, login);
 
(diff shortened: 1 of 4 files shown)


libathemecore/services.c
```diff
@@ -415,6 +415,8 @@ bool ircd_logout_or_kill(struct user *u, const char *login)
 	if (req.allowed)
```
ilbelkyr, line 415: i'd probably prefer a set of curly braces on the other `if` here, to make things more obvious – it's not required though. @aaronmdjones thoughts on style here?

ilbelkyr (commented):
so far this looks good. there's also another `ircd_on_logout` call in `myuser_rename`; that should probably be gated behind the same check (mirroring #830)?

PR #819, opened 2021-08-16
simplify NS INFO timestamp logic, indicate auspex being used

jesopo:
I think I'd like to extend this to other data too

### DIFF
--- a/modules/nickserv/info.c
+++ b/modules/nickserv/info.c
@@ -97,7 +97,7 @@ ns_cmd_info(struct sourceinfo *si, int parc, char *parv[])
 		return;
 	}
 
-	hide_info = (mu->flags & MU_PRIVATE) && (mu != si->smu) && !has_user_auspex;
+	hide_info = (mu->flags & MU_PRIVATE) && (mu != si->smu);
 
 	if (!nicksvs.no_nick_ownership)
 	{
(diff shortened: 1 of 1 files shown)


jesopo:
maybe there's a more succinct way to indicate this? given opers are expected to know what they're doing (not enough to notice `Private`...) it could be something relatively unobtrusive but slightly less clear? dunno

jesopo:
```
05:03:58 < dwfreed> jess: change hide_info to only test if user has private in flags, and then add a separate cached variable for whether the looker has auspex; then change the else to if !hide_info || auspex
05:04:19 < dwfreed> jess: (for #819)
```

edk0:
just to bikeshed a bit, the warning for hidden emails is `(hidden)`

jesopo:
this became a bigger change than I'd anticipated. this now significantly simplifies which timestamps are shown when such that `Last seen` and `User seen` are always shown even if redundant

ilbelkyr (changes requested):
I like this, both for making things more explicit in cases that have previously caused confusion and to make it clear when services operators are seeing private info.

There's an oversight in the implementation, skipping privilege checks when handling offline users, but once that's fixed I'm thinking this looks good to merge

ilbelkyr (approved):

PR #822, opened 2021-08-28
add hooks for adding and removing a certificate fingerprint

jesopo:
this is to support a module that will warn users when they add a certificate fingerprint that doesn't look like what we're expecting, i.e.:

```
<jess> cert add f10e2821bbbea527ea02200352313bc059445190
-NickServ- Warning: fingerprint of unexpected length. Expected hash algorithm: SHA512.
-NickServ- Added fingerprint f10e2821bbbea527ea02200352313bc059445190 to your fingerprint list.
```

### DIFF
--- a/include/atheme/hook.h
+++ b/include/atheme/hook.h
@@ -242,7 +242,7 @@ struct hook_user_certfp
 {
 	struct sourceinfo * si;
 	struct myuser *     mu;
-	const char *        certfp;
+	char                certfp[512];
 };
 
 void hook_del_hook(const char *, hook_fn);
--- a/include/atheme/hook.h
+++ b/include/atheme/hook.h
@@ -238,6 +238,13 @@ struct hook_user_req
 	struct mynick *     mn;
 };
 
+struct hook_user_certfp
+{
+	struct sourceinfo * si;
+	struct myuser *     mu;
+	const char *        certfp;
+};
+
 void hook_del_hook(const char *, hook_fn);
 void hook_add_hook(const char *, hook_fn);
 void hook_add_hook_first(const char *, hook_fn);
--- a/modules/nickserv/cert.c
+++ b/modules/nickserv/cert.c
@@ -134,7 +134,7 @@ ns_cmd_cert(struct sourceinfo *si, int parc, char *parv[])
 
 		struct hook_user_certfp hdata = {
 			.si = si,
-			.mu = mu
+			.mu = mu,
 		};
 		mowgli_strlcpy(hdata.certfp, mcfp, sizeof hdata.certfp);
 		hook_call_user_certfp_add(&hdata);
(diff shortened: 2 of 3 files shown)


jesopo:
something that's been pointed out is this can also be used to munge the certfp, to e.g. strip colons

include/atheme/hook.h
```diff
@@ -238,6 +238,13 @@ struct hook_user_req
 	struct mynick *     mn;
 };
 
+struct hook_user_certfp
+{
+	struct sourceinfo * si;
+	struct myuser *     mu;
+	char                certfp[512];
```
glguy, line 245: Instead of a fixed-sized allocation here, we could have this field be a `char *` to a `sstrdup`d copy of the fingerprint. If someone wanted to resize it up or down they'd be able to replace the pointer with a new one (making it bigger seems unlikely, but would work). We could free whatever happened to be in this field after calling the hook. (If a hook happened to replace it with a NULL and took ownership of what used to be there, no big deal)

ilbelkyr: That'd also work, but involves an extra heap allocation. I'm not sure there's much benefit to it in this case; practically, there's a length limit for fingerprints anyway for them to be usable via the IRC protocol framing (and in `CERT LIST` output).

`512` does feel like a bit of a magic number here, though I'm not sure there's a good constant defined for it – we don't specify a size limit for certfps, and `BUFSIZE` seems slightly excessive as the value needs to fit within (less than) 512 bytes for it to be usable in any case

modules/nickserv/cert.c
```diff
@@ -167,14 +175,23 @@ ns_cmd_cert(struct sourceinfo *si, int parc, char *parv[])
...
-		cert = mycertfp_find(parv[1]);
+
+		struct hook_user_certfp hdata = {
+			.si = si,
+			.mu = mu
+		};
+		mowgli_strlcpy(hdata.certfp, parv[1], sizeof hdata.certfp);
+		hook_call_user_certfp_del(&hdata);
+
+		cert = mycertfp_find(hdata.certfp);
```
ilbelkyr, line 186: If the `user_certfp_del` hook changed the certfp, we probably want to check for the unmodified certfp first, falling back to the modified certfp; otherwise, users may be unable to easily remove a fingerprint from their list if it is in a format the hook will modify.

(This issue can come up if the add hook doesn't match the delete hook, or more typically, if the user added a fingerprint at a time when the delete hook was not present.)

ilbelkyr (changes requested):

ilbelkyr:
additional consideration: for the hook on adding a certfp, should we allow the hook to reject the fingerprint entirely?

jesopo:
> additional consideration: for the hook on adding a certfp, should we allow the hook to reject the fingerprint entirely?

I guess it could be made a null pointer to indicate this?

(unrelated: I apologise that I seem to be reflexively adding newlines for aesthetic purposes)

include/atheme/hook.h
aaronmdjones: I'm fine with `512` here, and any registered hooks that wanted to modify the fingerprint would usually have to copy it to a temporary buffer and then write it back anyway, at which point they run into the same considerations.

PR #823, opened 2021-08-31
libathemecore/ptasks: handle status-prefixed channel messages

aaronmdjones:
If a channel message comes in with a status prefix, we should still treat it as a channel message.

This does not raise concerns about chanops hiding the usage of fantasy commands because they can always private message services directly and the output of potentially-abusive commands (for example, `KICK`) is signed off with their account name anyway (which will not change).

Under the solanum family of IRCds, there exists a channel mode `+z`, which will transform messages to `#channel` into messages to `@#channel` if the message would have been blocked by `+m`/`+b`/`+q`. This is to clearly distinguish to channel operators that the only reason they are seeing the message is that they are operators, and that other channel users will not have seen it.

ChanServ is also a channel operator, but it would not have received the message, because services ignored status-prefixed channel messages. This prevented users using fantasy commands if they are affected by `+z`; for example, by the channel being `+m`. Confusingly, this would have allowed "!deop" to work, but not "!op".

Reported by juggs on IRC, who did not respond to a request for the proper attribution in Reported-by.

### DIFF
--- a/libathemecore/ptasks.c
+++ b/libathemecore/ptasks.c
@@ -567,17 +567,34 @@ handle_message(struct sourceinfo *si, char *target, bool is_notice, char *messag
 	if (si->su == NULL)
 		return;
 
-	/* if this is a channel, handle it differently. */
-	if (*target == '#')
+	// If the target is a channel, handle it differently
+	if ((p = strchr(target, '#')) != NULL)
 	{
-		handle_channel_message(si, target, is_notice, message);
-		return;
+		/* Account for the fact that the message may have a status prefix on it; for example, a message
+		 * to @#channel should still be treated as a message to #channel rather than ignored.   -- amdj
+		 */
+		bool is_status_prefixed = false;
+
+		for (unsigned int i = 0; p == (target + 1) && prefix_mode_list && prefix_mode_list[i].mode; i++)
+		{
+			if (target[0] == prefix_mode_list[i].mode)
+			{
+				is_status_prefixed = true;
+				break;
+			}
+		}
+
+		if (p == target || is_status_prefixed)
+		{
+			(void) handle_channel_message(si, p, is_notice, message);
+			return;
+		}
 	}
 
 	target_u = NULL;
 	si->service = NULL;
-	p = strchr(target, '@');
-	if (p != NULL)
+
+	if ((p = strchr(target, '@')) != NULL)
 	{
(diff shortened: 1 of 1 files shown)


kaniini:
Looks fine to me.

libathemecore/ptasks.c
```diff
@@ -567,17 +567,34 @@ handle_message(struct sourceinfo *si, char *target, bool is_notice, char *messag
...
+	if ((p = strchr(target, '#')) != NULL)
 	{
-		handle_channel_message(si, target, is_notice, message);
-		return;
+		/* Account for the fact that the message may have a status prefix on it; for example, a message
+		 * to @#channel should still be treated as a message to #channel rather than ignored.   -- amdj
+		 */
+		bool is_status_prefixed = false;
+
+		for (unsigned int i = 0; p == (target + 1) && prefix_mode_list && prefix_mode_list[i].mode; i++)
```
jesopo, line 578: i feel like the inclusion of static logic in a for condition is needlessly abstruce. this took me 5 minutes to understand

jesopo: is there much at risk to just do

```c
if ((p = strchr(target, '#')) != NULL)
{
    (void) handle_channel_message(si, p, is_notice, message);
    return;
}
```

?

it feels like having an octothorpe in a target name when it's not ultimately a channel is a protocol framing issue

kaniini: This was proposed about a decade ago, but there was something wrong with it.  I forget what it was now, though.

Honestly, it seems fine, we were probably bikeshedding about a non-practical issue.

aaronmdjones: > is there much at risk to just do
> 
> ```c
> if ((p = strchr(target, '#')) != NULL)
> {
>     (void) handle_channel_message(si, p, is_notice, message);
>     return;
> }
> ```

Think of charybdis-family IRCds where you send a PRIVMSG to `$#some.mask` and `#some.mask` happens to be an extant channel.

ilbelkyr:
hm, for solanum (via ts6-generic) we don't currently advertise support for `CHW` so simply won't receive statusmsg in the first place afaict; we'll want to start advertising that if this is merged

we _do_ advertise `EOPMOD`, which appears to be a no-op without `CHW`; if we keep advertising that, we'll be sent messages affected by `+z` to a `=#channel` target which we'd need to specifically support. @aaronmdjones thoughts on that bit?

PR #824, opened 2021-09-01
allow config file sopers to be specified by ?UID

jesopo:
and also allow config uid sopers to change accountname

### DIFF
--- a/libathemecore/privs.c
+++ b/libathemecore/privs.c
@@ -265,6 +265,25 @@ soper_find_named(const char *name)
 
 	return NULL;
 }
+struct soper *
+soper_find_eid(const char *eid)
+{
+	struct soper *soper;
+	mowgli_node_t *n;
+	char lookup_eid[IDLEN+2]; // 2 for ? and \0
+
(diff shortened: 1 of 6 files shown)


libathemecore/privs.c
```diff
@@ -265,6 +265,25 @@ soper_find_named(const char *name)
 
 	return NULL;
 }
+struct soper *
+soper_find_uid(const char *uid)
+{
+	struct soper *soper;
+	mowgli_node_t *n;
+	char lookup_uid[UIDLEN+1];
```
edk0, line 273: I think this should be IDLEN+2

jesopo: how come

jesopo: oh null termination i forgot about that

edk0: 2 for ? and null terminator, IDLEN because UIDLEN is for s2s UIDs

jesopo:
I think this will currently not let you configure DB sopers by `?ID`. do we want that?

PR #826, opened 2021-09-02
services.c/enforce.c: honor no login limit flag

### LINKED ISSUE #825: CertFP logins do not honor user:nologinlimit priv
https://github.com/atheme/atheme/blob/98ef83cfdf1bc6c86025bbdbd69354ae2a957a37/libathemecore/services.c#L635

examknow:
Closes #825

### DIFF
--- a/modules/nickserv/enforce.c
+++ b/modules/nickserv/enforce.c
@@ -445,7 +445,7 @@ ns_cmd_regain(struct sourceinfo *si, int parc, char *parv[])
 	}
 	if ((si->smu == mn->owner) || verify_password(mn->owner, password))
 	{
-		if (MOWGLI_LIST_LENGTH(&mn->owner->logins) >= me.maxlogins)
+		if (user_loginmaxed(mn->owner))
 		{
 			command_fail(si, fault_toomany, _("You were not logged in."));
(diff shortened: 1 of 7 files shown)


jesopo:
I wanted to try instead refactoring this check so it's not duplicated everywhere

spb:
As @jesopo said, this is the wrong way to fix it. The problem stems from the logic being duplicated, so that's what needs changing.

examknow:
> 
> 
> As @jesopo said, this is the wrong way to fix it. The problem stems from the logic being duplicated, so that's what needs changing.

I agree and will be updating the PR with those changes, hence the draft

jesopo (approved):

modules/nickserv/enforce.c
```diff
@@ -445,7 +445,7 @@ ns_cmd_regain(struct sourceinfo *si, int parc, char *parv[])
 	}
 	if ((si->smu == mn->owner) || verify_password(mn->owner, password))
 	{
-		if (MOWGLI_LIST_LENGTH(&mn->owner->logins) >= me.maxlogins)
+		if (user_loginmaxed(si->smu))
```
glguy, line 448: Shouldn't this be testing `mn->owner` instead of `si->smu->owner` because while in the first case (`si->smu == mn->owner`) these are equal in the second case they won't be (`verify_password(mn->owner, password)`)?

```suggestion
		if (user_loginmaxed(mn->owner))
```

glguy (approved):

jesopo (approved):
tested locally. works

PR #827, opened 2021-09-03
check user_can_login *before* checking certfp

jesopo:
this isn't important, but you can actually expose that your current certfp matches an account without being able to log in to it if the target account already has max logins


libathemecore/services.c
```diff
@@ -617,6 +617,15 @@ handle_certfp(struct sourceinfo *si, struct user *u, const char *certfp)
 	if (u->myuser != NULL)
 		return;
 
+	req.si = si;
+	req.mu = mu;
```
ilbelkyr, line 621: `mu` is uninitialized at this point

ilbelkyr (changes requested):

libathemecore/services.c
jesopo: can we just go with expecting `mu` to be NULL sometimes? unsure what that'd break

dwfreed: `sasl_blacklist` in contrib needs a valid `mu` for logging purposes (it prints the account name they attempted to log into).  This is thus far the only consumer of `user_can_login` in atheme or contrib.

jesopo: see https://github.com/atheme/atheme-contrib-modules/pull/53, and incoming commit to this PR

jesopo:
if this solution is acceptable, merge atheme/atheme-contrib-modules#53 first (and get e.g. libera.chat to deploy these at the same time)

libathemecore/services.c
```diff
@@ -617,10 +617,20 @@ handle_certfp(struct sourceinfo *si, struct user *u, const char *certfp)
 	if (u->myuser != NULL)
 		return;
 
-	if ((mcfp = mycertfp_find(certfp)) == NULL)
+	if ((mcfp = mycertfp_find(certfp)) != NULL)
+		mu = req.mu = mcfp->mu;
+
+	req.si = si;
```
aaronmdjones, line 623: This needs to initialise `req.mu` with a value, or its value will be undefined in the hook functions that access it.

You can do this by initialising `mu` to `NULL` at the top of this function, and then adding `req.mu = mu;` here, taking the `req.mu` assignment out of the conditional block above.

aaronmdjones (changes requested):

PR #829, opened 2021-09-15
add a hook to allow modules to handle sending emails

jesopo:
placement of the hook call may be controversial, returning `-1` to avoid falsiness but not get in the way of exit codes may also be controversial

### DIFF
--- a/libathemecore/email.c
+++ b/libathemecore/email.c
@@ -317,6 +317,17 @@ sendemail(struct user *u, struct myuser *mu, const char *type, const char *email
 			u->nick, u->user, u->vhost, u->ip ? u->ip : u->host,
 			type, entity(mu)->name, email);
 
+	struct hook_sendemail_req hdata = {
+		.u     = u,
+		.mu    = mu,
+		.type  = type,
+		.email = email,
+		.param = param
+	};
+	hook_call_sendemail(&hdata);
+	if (hdata.handled)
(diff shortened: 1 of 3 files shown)


libathemecore/email.c
```diff
@@ -317,6 +317,17 @@ sendemail(struct user *u, struct myuser *mu, const char *type, const char *email
 			u->nick, u->user, u->vhost, u->ip ? u->ip : u->host,
 			type, entity(mu)->name, email);
 
+	struct hook_sendemail_req hdata = {
```
glguy, line 320: I know that `.handled` will default initialize to `false` here, but I think it'd be nice to be explicit about that as we're initializing all the other fields.

libathemecore/email.c
```diff
@@ -317,6 +317,17 @@ sendemail(struct user *u, struct myuser *mu, const char *type, const char *email
...
+		.u     = u,
+		.mu    = mu,
+		.type  = type,
+		.email = email,
+		.param = param
+	};
+	hook_call_sendemail(&hdata);
+	if (hdata.handled)
+		return -1;
```
jillest, line 329: `in` is not closed here.

jillest (commented):
Depending on what you need this for, it may make more sense to use a plain function pointer rather than hooks (like protocol modules).

PR #830, opened 2021-09-20
don't do ircd_on_login in myuser_rename if we're MU_WAITAUTH

glguy:


libathemecore/account.c
```diff
@@ -371,7 +371,7 @@ myuser_rename(struct myuser *mu, const char *name)
 	entity(mu)->name = newname;
 
 	myentity_put(entity(mu));
-	if (authservice_loaded)
+	if (authservice_loaded && !(mu->flags & MU_WAITAUTH))
```
ilbelkyr, line 374: hrm, i'm not sure the `MU_WAITAUTH` check here ought to cover the hook call as well?

ilbelkyr: (nevermind, misread)

PR #840, opened 2021-11-10
make MARK DEL RESTORED work for unregistered accounts

jesopo:


modules/nickserv/multimark.c
```diff
@@ -1032,6 +1046,22 @@ ns_cmd_multimark(struct sourceinfo *si, int parc, char *parv[])
 				break;
 		}
 
+		MOWGLI_ITER_FOREACH_SAFE(n, tn, rl->head)
+		{
+	                rm = n->data;
```
JeDaYoshi, line 1051: What happened with the indentation in there?

jesopo: oops, thanks!

PR #841, opened 2021-11-10
tell anything logged in to an account when it has been (F)VERIFYed 

jesopo:


modules/nickserv/verify.c
```diff
@@ -190,6 +190,9 @@ ns_cmd_fverify(struct sourceinfo *si, int parc, char *parv[])
 		metadata_delete(mu, "private:verify:emailchg:newemail");
 		metadata_delete(mu, "private:verify:emailchg:timestamp");
 
+		// tell the user they have been verified
+		myuser_notice(nicksvs.nick, mu, _("\2%s\2 has now been verified."), entity(mu)->name);
```
jesopo, line 194: not 100% on the wording here. might not make as much sense when it's not directly after you've personally executed a verification command

jesopo:
I'd like to extend this to `VERIFY` too, not just `FVERIFY`

ilbelkyr:
hmm, we'll also send the "thank you for verifying…" message to whoever ran the command, which won't make much sense when that's actually someone running fverify. Could add a parameter to `ns_verify_activate_account` to indicate normal vs `fverify`, though 99% of use cases would already be covered by checking `si->smu == mu` (but it *is* technically possible to fverify your own account if you already have privs)

modules/nickserv/verify.c
ilbelkyr: agreed, i'd still send a message but indicate something like "\2%s\2 has been verified by network staff" or similar

in the future, we might want this to allow off-IRC email verifications (e.g. via xmlrpc or some other api) as that's something that people have expressed interest in, but imo that's something to consider once we add an API for that (calling fverify via atheme.command is imo not a great way to do it)

ilbelkyr (approved):

PR #847, opened 2021-11-28
allow hooks to force account expire

jesopo:
~currently untested~


jesopo:
maybe todo: similar work for `hook_call_nick_check_expire`?

jesopo:
tested. seems fine to me

PR #850, opened 2021-12-29
nickserv/enforce: REGAIN: fix race condition

ilbelkyr:
FNC semantics with most protocols are that services request the nick change, but the actual nick change itself has to originate from the user's own server. This means that we must wait for the Guest nick change to be applied before changing the regaining user's nick, as otherwise that change might be applied first on part of the network, requiring an ircd-side resolution of the nick collision resulting from this as the original person using the nick has not been renamed yet.

This probably still has a number of imperfections; it's essentially a rebase of a handful of commits from ~~almost two years ago~~ even longer ago by now and hasn't seen much real-world testing yet.

as @dwfreed pointed out, this has been sitting on a branch for ages now. potential considerations:

* do we want a timeout in case of network issues causing the target's server to be unresponsive, so users can still regain nicks even if their target's server is dying/splitting?

### DIFF
--- a/modules/nickserv/enforce.c
+++ b/modules/nickserv/enforce.c
@@ -25,12 +25,22 @@ struct enforce_timeout
 	mowgli_node_t node;
 };
 
+struct pending_regain
+{
+	char nick[NICKLEN + 1];
+	stringref client;
+	time_t client_ts;
+};
+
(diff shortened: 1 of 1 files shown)


delthas:
Any chance of getting this rebased & merged? :smile:

PR #853, opened 2022-01-13
Support +T (CMODE_NONOTICE)

skizzerz:
This was already defined in the charybdis.h file, so just need to make
the rest of atheme recognize the mode.


skizzerz:
it's a proven fact that wolves cannot read

PR #862, opened 2022-03-28
don't do_sethost_all in NS VHOST for unverified accounts

jesopo:


glguy:
It looks like even with this patch someone could reconnect and the user_identify hook would then set the vhost for them. 

Setting a vhost on someone who hasn't finished registration might be something an soper wanted to do for some reason. Maybe we shouldn't prevent it?

If we do want to prevent it would the vhost be applied in this case after verification was complete?

PR #868, opened 2022-04-30
log account name, rather than provided target, in multimark

jesopo:
(apologies for changing some stuff to `const` while i was in the neighbourhood)

### DIFF
--- a/modules/nickserv/multimark.c
+++ b/modules/nickserv/multimark.c
@@ -725,9 +725,9 @@ multimark_needforce(struct hook_user_needforce *hdata)
 static void
 ns_cmd_multimark(struct sourceinfo *si, int parc, char *parv[])
 {
-	char *target = parv[0];
-	char *action = parv[1];
-	char *info = parv[2];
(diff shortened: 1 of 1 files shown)


jesopo:
the problem being solved here is
```
/ns mark =person add bad person
```
logs as
```
<@NickServ> jess MARK:ADD: =person bad person
```
which is not great for log grepping

modules/nickserv/multimark.c
```diff
@@ -772,6 +772,7 @@ ns_cmd_multimark(struct sourceinfo *si, int parc, char *parv[])
 		command_fail(si, fault_nosuch_target, STR_IS_NOT_REGISTERED, target);
 		return;
 	}
+	target = entity(mu)->name;
```
dwfreed, line 775: This will null-deref if you want to list the marks of a dropped account

dwfreed (changes requested):

aaronmdjones (approved):

PR #870, opened 2022-05-15
print badmail reason in badmail hit log line

jesopo:


jesopo:
i wonder if maybe we should also include the badmail entry that caused this badmail to hit

alyx (approved):

alyx (approved):

aaronmdjones (approved):

PR #875, opened 2022-06-15
LG_VERBOSE log new account entity IDs

jesopo:


jesopo:
might make more sense as `LG_INFO`?

jesopo:
i've got a better idea for how to do this (verbose state streams)

PR #876, opened 2022-06-24
chanserv: allow hiding AKICK entries from FLAGS

phy1729:


jesopo:
the usecase here is that `CS FLAGS #foo` will become hard to manage if a channel has a lot of AKICKs, and there's already `CS AKICK #foo LIST` for it

PR #879, opened 2022-07-16
Suggest searching for network name on nick reg

Kufat:
Because this one comes up frequently on my network.


modules/nickserv/register.c
```diff
@@ -185,6 +185,7 @@ ns_cmd_register(struct sourceinfo *si, int parc, char *parv[])
 
 		command_success_nodata(si, _("An email containing nickname activation instructions has been sent to \2%s\2."), mu->email);
 		command_success_nodata(si, _("Please check the address if you don't receive it. If it is incorrect, DROP then REGISTER again."));
+		command_success_nodata(si, _("Your mail provider may filter the message as spam or bulk mail. Try searching for '%s'."), me.netname);
```
jillest, line 188: Existing convention is to make this kind of variables bold instead of putting quotes around them.

```suggestion
		command_success_nodata(si, _("Your mail provider may filter the message as spam or bulk mail. Try searching for \2%s\2."), me.netname);
```

PR #880, opened 2022-07-20
refactor out hostmask formatting

jesopo:
i plan to use this function in more places that already format hostmasks, but i wanted to open this PR early to see if this is the right way of going about this. returning a pointer means (i believe) that you can't call this and store its results consecutively, but i think that might be an ok cost for not having to free things

### DIFF
--- a/libathemecore/users.c
+++ b/libathemecore/users.c
@@ -21,6 +21,8 @@ static mowgli_heap_t *user_heap = NULL;
 mowgli_patricia_t *userlist;
 mowgli_patricia_t *uidlist;
 
+char hostmaskbuf[NICKLEN + USERLEN + HOSTLEN + 2];
+
 static void
 user_delete_cb(void *const restrict user)
 {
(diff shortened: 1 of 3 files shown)


libathemecore/users.c
```diff
@@ -21,6 +21,8 @@ static mowgli_heap_t *user_heap = NULL;
 mowgli_patricia_t *userlist;
 mowgli_patricia_t *uidlist;
 
+char hostmaskbuf[NICKLEN + USERLEN + HOSTLEN + 2];
```
ilbelkyr, line 24: this needs an extra byte for the string terminator; i'd also add a comment explaining where the number comes from to avoid people making mistakes in the future

you could also put this as `static` inside the `format_hostmask` function i think; might be a bit cleaner?

libathemecore/users.c
```diff
@@ -673,6 +675,13 @@ find_user_banned_channel(struct user *const restrict u, const char ban_type)
 	return NULL;
 }
 
+const char *
+format_hostmask(const struct user *user, const enum host_type type)
+{
+	sprintf(hostmaskbuf, "%s!%s!%s", user->nick, user->user, user->vhost);
```
ilbelkyr, line 681: this currently just does `user->vhost` without checking the requested host type

i'd probably use `snprintf` with `sizeof hostmaskbuf` here for safety

ilbelkyr (changes requested):
moving code like this into helper functions is an idea i'm in favor of, i think, especially if it's going to be reused more

as for returning the same buffer each time, if the caller needs to store it they can just `sstrdup()` it i'd say

PR #891, opened 2022-12-27
added documentation for OFF

Svetlana-T:
Credit to KindOne at #libera for mentioning this is not documented

### DIFF
--- a/help/default/cservice/set_mlock
+++ b/help/default/cservice/set_mlock
@@ -8,6 +8,11 @@ recreating the channel (this includes keys).
 
 Syntax: SET <#channel> MLOCK [modes]
 
+The MLOCK can also be removed, so that no modes are 
+enforced or locked.
+
+Syntax: SET <#channel> MLOCK OFF
+
 Examples: (some may use modes your ircd does not support)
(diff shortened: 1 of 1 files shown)


RealKindOne:
After reviewing the chanserv/set_mlock source code and a bit of testing I have determined that I am wrong on that command.

You only need ```cs set #channel mlock``` remove the mlock.  Anything invalid like ```OFF```, nonexistent modes, oper-only modes (while not oper), has the same effect and remove the mlock.

There are a few other command's that accept multiple ways to disable them, but they are not documented on how:

```
ChanServ SET EMAIL
ChanServ SET ENTRYMSG 
ChanServ SET URL
```
They accept ``` ```, ```NONE```, and ```OFF```


https://github.com/atheme/atheme/blob/ab7319d7868b8f80540850c903c4ea6c3268b894/modules/chanserv/set_entrymsg.c#L40


https://github.com/atheme/atheme/blob/ab7319d7868b8f80540850c903c4ea6c3268b894/modules/chanserv/set_email.c#L39

https://github.com/atheme/atheme/blob/ab7319d7868b8f80540850c903c4ea6c3268b894/modules/chanserv/set_url.c#L39

PR #894, opened 2023-01-29
This lets Unreal know about the package and version

ValwareIRC:
This lets the package name be declared in a way where unrealircd's new admin panel can display


aaronmdjones:
Based on my impressions from https://github.com/unrealircd/unrealircd/blob/2417ed0a46ccf3f29e86046ba231fd86dcab25bd/src/modules/protoctl.c#L221-L282 -- appears to work as described.

PR #897, opened 2023-03-22
unreal4: CHGIDENT and CHGHOST on SASL

syzop:
This makes it so in the unreal4 protocol module `CHGIDENT` and `CHGHOST` get sent on SASL when instructed to do so.

Actually this doesn't help much, since right now the `svslogin_sts` function is always called with `*` for both `user` and `host` but that is another issue :D

### DIFF
--- a/modules/protocol/unreal4.c
+++ b/modules/protocol/unreal4.c
@@ -708,6 +708,10 @@ unreal_svslogin_sts(const char *target, const char *nick, const char *user, cons
 		*p = '\0';
 
 	sts(":%s SVSLOGIN %s %s %s", saslserv->me->nick, servermask, target, entity(account)->name);
+	if (strcmp(user, "*"))
+		sts(":%s CHGIDENT %s %s", me.numeric, target, user);
+	if (strcmp(host, "*"))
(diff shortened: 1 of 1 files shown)


syzop:
So about the other issue.... there's only one place where svslogin_sts is called and that is in `modules/saslserv/main.c`.
Sadly it always calls it with `*` arguments for `user` and `host`:
```c
(void) svslogin_sts(p->uid, "*", "*", cloak, mu);
```
I have close to zero experience with fiddling with atheme code, but if this would be enhanced to include/set the ident and host in case there is a vident/vhost, that would be great. It would not only fix this issue for UnrealIRCd (with this PR) but a quick look in the code says it should also fix inspircd which has the same in `inspircd_svslogin_sts`.

The purpose of all this, is that the vident/vhost is set directly during SASL, before the user actually comes online. That way the ident/host is already set correctly on the user before the user starts joining channels (either automatically or due to rejoin). No longer a race condition where the client is joined to some or all channels and then their host is changed, prematurely exposing their original host and/or causing things like "rejoining due to user@host change" and userhost notifications to be sent.

PR #898, opened 2023-04-02
modules/operserv/: add a channel join rate monitoring module

aaronmdjones:
I thought this was better as a first-party module than to put it in contrib. I have reworked it to adhere to the 7.3 type system and conventions, and add module reload support such that no data loss will occur.

This still needs to be thoroughly tested.

### DIFF
--- a/modules/operserv/joinrate.c
+++ b/modules/operserv/joinrate.c
@@ -1,0 +1,392 @@
+/*
+ * SPDX-License-Identifier: ISC
+ * SPDX-URL: https://spdx.org/licenses/ISC.html
+ *
+ * Copyright (C) 2021 Jess <github@lolnerd.net>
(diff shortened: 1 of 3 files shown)


modules/operserv/joinrate.c
```diff
@@ -0,0 +1,392 @@
...
+	// Is there a token left for us to consume?
+	if (instance->tokens > 0)
+	{
+		instance->tokens--;
+	}
+	else if ((CURRTIME - instance->last_warn_time) >= 30)
+	{
+		instance->last_warn_time = CURRTIME;
+
+		(void) slog(LG_INFO, "JOINRATE: \2%s\2 exceeds warning threshold (%d joins in %ds)",
```
jesopo, line 144: this log line undercounts `chrate` by 1

jesopo (approved):

PR #899, opened 2023-04-06
include FNC_REGAIN/FNC_FORCE in ts6-generic's RSFNC

jesopo:
it may be better to only send this extra parameter to servers whose `CAPAB`/`GCAP` have indicated they support it. thoughts?


jesopo:
c.f. https://github.com/solanum-ircd/solanum/pull/406

jesopo:
> it may be better to only send this extra parameter to servers whose `CAPAB`/`GCAP` have indicated they support it. thoughts?

annoyingly, it doesn't look like we record `GCAP` so we can't (currently) check what capabs the target server has, given `RSFNC` travels through uplink to target server

PR #901, opened 2023-04-15
nickserv: allow account emails to be made optional

examknow:
This change allows users to not associate an email with their account by passing `noemail` to the `email` parameter on NickServ's `REGISTER` and `SET EMAIL` commands. Networks must have `email_optional` in their NickServ config for this to work.


include/atheme/services.h
```diff
@@ -50,6 +50,7 @@ struct nicksvs
 	bool            spam;
 	bool            no_nick_ownership;
 	bool		bad_password_message;
+	bool		email_optional;
```
dwfreed, line 53: This is a breaking ABI change. It should be added to the end instead.

modules/nickserv/register.c
```diff
@@ -83,7 +83,7 @@ ns_cmd_register(struct sourceinfo *si, int parc, char *parv[])
 		return;
 	}
 
-	if (!validemail(email))
+	if (!validemail(email) && !(nicksvs.email_optional && !strcasecmp(email, "noemail")))
```
dwfreed, line 86:
```suggestion
	if (!validemail(email) && !(nicksvs.email_optional && strcasecmp(email, "noemail") == 0))
```

modules/nickserv/register.c
```diff
@@ -167,7 +167,7 @@ ns_cmd_register(struct sourceinfo *si, int parc, char *parv[])
 		}
 	}
 
-	if (me.auth == AUTH_EMAIL)
+	if (me.auth == AUTH_EMAIL && strcasecmp(email, "noemail"))
```
dwfreed, line 170:
```suggestion
	if (me.auth == AUTH_EMAIL && strcasecmp(email, "noemail") != 0)
```

modules/nickserv/set_email.c
```diff
@@ -41,7 +41,7 @@ ns_cmd_set_email(struct sourceinfo *si, int parc, char *parv[])
 		return;
 	}
 
-	if (!validemail(email))
+	if (!validemail(email) && !(nicksvs.email_optional && !strcasecmp(email, "noemail")))
```
dwfreed, line 44:
```suggestion
	if (!validemail(email) && !(nicksvs.email_optional && strcasecmp(email, "noemail") == 0))
```

modules/nickserv/set_email.c
```diff
@@ -53,7 +53,7 @@ ns_cmd_set_email(struct sourceinfo *si, int parc, char *parv[])
 		return;
 	}
 
-	if (me.auth == AUTH_EMAIL)
+	if (me.auth == AUTH_EMAIL && strcasecmp(email, "noemail"))
```
dwfreed, line 56:
```suggestion
	if (me.auth == AUTH_EMAIL && strcasecmp(email, "noemail") != 0)
```

dwfreed (changes requested):

dwfreed (approved):

modules/nickserv/register.c
```diff
@@ -83,13 +83,13 @@ ns_cmd_register(struct sourceinfo *si, int parc, char *parv[])
 		return;
 	}
 
-	if (!validemail(email))
```
aaronmdjones, line 86: It would be better to indent both conditional blocks and test `! (nicksvs.email_optional && strcasecmp(email, "noemail") == 0)` only once. This would also avoid calling the rather expensive function `email_within_limits()` on garbage data.

You also need to take care of excluding `MU_EMAILMEMOS` from `defuflags` around line 147 if the user has passed `noemail`; otherwise, they will still have `MU_EMAILMEMOS` if the services admin has configured new accounts to have that flag by default.

modules/nickserv/set_email.c
```diff
@@ -41,19 +41,19 @@ ns_cmd_set_email(struct sourceinfo *si, int parc, char *parv[])
 		return;
 	}
 
-	if (!validemail(email))
```
aaronmdjones, line 44: Same here (regarding testing the condition only once).

modules/nickserv/set_email.c
```diff
@@ -77,6 +77,14 @@ ns_cmd_set_email(struct sourceinfo *si, int parc, char *parv[])
 		return;
 	}
 
+	if (strcasecmp(email, "noemail") == 0 && (si->smu->flags & MU_EMAILMEMOS))
+	{
+		// MU_EMAILMEMOS cannot be used with noemail
+		// if they have it, we'll take it away and log that we've done so
+		logcommand(si, CMDLOG_SET, "SET:EMAILMEMOS:OFF (user is noemail)");
```
aaronmdjones, line 84: There's no need to log this.

aaronmdjones (changes requested):
You are missing equivalent `validemail()` checks in `modules/nickserv/sendpass` and `modules/nickserv/sendpass_user`.

PR #909, opened 2023-08-27
Update unreal4.c: SVSLOGIN (logout) - Tells atheme when a user was lo…

ValwareIRC:
…gged out server-side

Tested with latest packaged version.


ValwareIRC:
This in conjunction with https://github.com/unrealircd/unrealircd/pull/231

PR #920, opened 2024-02-21
make vhost "looks like a CIDR" check more precise

jesopo:
i don't know if this code was intentionally very imprecise but it's blocking a cloak we'd like to use that is quite obviously fine


libathemecore/services.c
```diff
@@ -1246,8 +1246,13 @@ check_vhost_validity(struct sourceinfo *si, const char *host)
...
 	}
-	p = strrchr(host, '/');
-	if (p != NULL && isdigit((unsigned char)p[1]))
+	p = p2 = strrchr(host, '/');
+	// walk the vhost until we hit either the end or a non-digit
+	while (p != NULL && isdigit((unsigned char)p[1]))
+		p++;
+	// if we found a /, and are now at the end of the vhost, that
+	// means we found no non-digits, so it looks too much like a CIDR
+	if (p2 != NULL && p[1] == '\0')
```
dwfreed, line 1255: This doesn't need storing a second copy of the `strrchr` return value.  If the return value was `NULL` then `p` will still be `NULL` here; if the return value was not `NULL` then `p` will never be NULL either.

dwfreed (changes requested):

jesopo:
this PR may be a no-go anyway as it'd apparently require us to audit all the IRCds atheme can link with

PR #931, opened 2024-11-23
Update InspIRCd protocol module to use the 1205 (v3) protocol

### LINKED ISSUE #904: Please update protocol/inspircd to the 1205 (v3) protocol
The protocol/inspircd module is currently using the 1202 (v2) protocol. This works with v3 because of the v2 compatibility layer but when v4 is released at the end of the year it will no longer be able to link as the v2 compatibility layer has been removed. In order to work with v4 Atheme will need to migrate to the v3 (1205) protocol which works on v3 and will work on v4 through the compatibility layer.

There's a changelog between server versions [here](https://docs.inspircd.org/server/change-log/#1205-v3) and full documentation for each server message [here](https://docs.inspircd.org/server/messages/#messages).

SadieCat:
This drops support for v2 (which has been eol for years now) and adds support for v4 (via the v3 compatibility layer).

I'd appreciate help testing this. I have tested it myself but I might have missed some stuff.

Closes #904.

### DIFF
--- a/dist/atheme.conf.example
+++ b/dist/atheme.conf.example
@@ -119,7 +119,7 @@
  * Charybdis IRCd                               protocol/charybdis
  * ChatIRCd                                     protocol/chatircd1.1
  * DreamForge 4.6.7 or later                    protocol/dreamforge
- * InspIRCd 2.x and 3.x                         protocol/inspircd
+ * InspIRCd 3.x and 4.x                         protocol/inspircd
  * ircd-ratbox 2.0 and later                    protocol/ratbox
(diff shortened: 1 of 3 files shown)


alyx:
I've also tested it lightly, although "a server of me, myself and I having some great convos with ourself" is probably missing a couple of things :P 

Code-wise, looks good to me. Haven't seen anything blow up in testing, I think it's good :)

alyx:
Update from IRC: Might be broken on networks with multiple servers.

Will review.

wpbirney:
https://github.com/atheme/atheme/pull/933 resolves the issue i was having with multiple servers. My bad i didn't notice it during testing

PR #933, opened 2024-11-27
Fix parsing remote SERVER messages from InspIRCd.

SadieCat:
Fixes the issue reported by blackbeard on IRC.


wpbirney:
blackbeard here, this PR resolves that issue

PR #936, opened 2025-01-02
Fix various minor issues with the InspIRCd protocol module

SadieCat:
1. Topic changes were not handled correctly when sent by a user outside of the initial burst.

2. When logging "all" to !snotices a SNONOTICE message would be sent during a connection phase when it is not allowed.

3. Some obsolete code in m_ping was allowing the burst to end but not allowing the uplink to be considered synchronised.

4. When an operator on a non-directly connected server joined the channel using `/OJOIN` their channel prefix mode was not being mapped to a prefix character correctly.

5. CBANs should be added through ADDLINE not through the client command.

6. Account identifiers should be synchronised with the IRC server if public. This is used by the server for tracking users if the account is renamed.

### DIFF
--- a/modules/protocol/inspircd.c
+++ b/modules/protocol/inspircd.c
@@ -624,12 +624,14 @@ inspircd_ping_sts(void)
 static void
 inspircd_on_login(struct user *u, struct myuser *mu, const char *wantedhost)
 {
+	sts(":%s METADATA %s accountid :%s", me.numeric, u->uid, config_options.show_entity_id ? entity(mu)->id : "");
 	sts(":%s METADATA %s accountname :%s", me.numeric, u->uid, entity(mu)->name);
 }
 
 static bool
 inspircd_on_logout(struct user *u, const char *account)
 {
+	sts(":%s METADATA %s accountid :", me.numeric, u->uid);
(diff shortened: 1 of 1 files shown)


modules/protocol/inspircd.c
```diff
@@ -385,6 +385,12 @@ inspircd_quit_sts(struct user *u, const char *reason)
 static void
 inspircd_wallops_sts(const char *text)
 {
+	if (!me.bursting)
+	{
+		slog(LG_DEBUG, "inspircd_wallops_sts(): refusing to send SNONOTICE while bursting: %s", text);
```
dwfreed, line 390: You should not log here, this could trigger an infinite loop if debug is sent to notices.

ilbelkyr: I think there's a check in `vslog_ext` nowadays that should make it no-op when called recursively, so this is probably fine (but the log message might get dropped if we got here by trying to log something else)

ilbelkyr: (that said, is this check supposed to be negated?)

SadieCat: This should be resolved now.

PR #939, opened 2025-01-26
Reject passwords that cannot be encrypted

aaronmdjones:
At the moment, if services fails to encrypt a password, it will fall back to storing it in plain text in the database. This is clearly undesirable. This can happen for a couple of reasons off of the top of my head:

- No encryption-capable crypto providers are loaded.

  This is extremely negligent, but is still possible. Services complains loudly every time it attempts to encrypt a password in this configuration, but that will not stop some negligent sysadmins from ignoring these messages and continuing to run it in this configuration.

- pbkdf2v2 is the only encryption-capable crypto provider loaded, it is configured in SCRAM mode with GNU libidn for SASLprep normalisation, and the user supplies a password containing invalid UTF-8 byte sequences or ASCII control characters.

- argon2 is the only encryption-capable crypto provider loaded, and it is configured to use far more memory than the system has available; high enough to cause libc to fall back to a direct mmap(2), and high enough for the kernel's overcommit logic to reject it as extreme, returning NULL.

This is because set_password() has no way to indicate to its caller whether it was able to encrypt the password or not.

This commit forces set_password() to encrypt the password or return failure to its caller. The callers now check whether it was successful and abort their operation if it was not.

Also add a note to the SASL SCRAM documentation mentioning the prohibited password characters, and the same to the help files for commands that take passwords (but only if the SASL SCRAM module is loaded).


aaronmdjones:
The following test was performed with PBKDF2v2 in SCRAM mode being the only loaded crypto provider:

```
Normal password (0x68 0x75 0x6e 0x74 0x65 0x72 0x32):

[2025-01-26 13:07:08] -> :42XAAAAAB PRIVMSG 00AAAAAAC :SET PASSWORD hunter2
[2025-01-26 13:07:08] crypt_password: encrypted password with provider 'crypto/pbkdf2v2'
[2025-01-26 13:07:08] NickServ amdj/AAAAAAAAB:amdj!~aaron@asteria.home.aaronmdjones.net[0::1] SET:PASSWORD
[2025-01-26 13:07:08] <- :00AAAAAAC NOTICE 42XAAAAAB :The password for amdj has been successfully changed.
```

```
Green IRC coloured password (0x03 0x33 0x68 0x75 0x6e 0x74 0x65 0x72 0x32 0x03):

[2025-01-26 13:07:12] -> :42XAAAAAB PRIVMSG 00AAAAAAC :SET PASSWORD hunter2
[2025-01-26 13:07:12] atheme_pbkdf2v2_scram_normalize: Prohibited code points in input
[2025-01-26 13:07:12] atheme_pbkdf2v2_compute: SASLprep normalization of password failed
[2025-01-26 13:07:12] crypt_password: ci->crypt() failed for provider 'crypto/pbkdf2v2'
[2025-01-26 13:07:12] crypt_password: all encryption-capable crypto providers failed
[2025-01-26 13:07:12] set_password: failed to encrypt password for account 'amdj'
[2025-01-26 13:07:12] <- :00AAAAAAC NOTICE 42XAAAAAB :There was an error setting your password. Please check it for any invalid characters and contact network staff if the issue persists.
```

invpw

aaronmdjones:
The following test was performed with Argon2 being the only loaded crypto provider, configured with a memory cost of 29 (512 GiB of RAM) on a machine with only 128 GiB:

```
[2025-01-26 14:07:58] -> :42XAAAAAC PRIVMSG 00AAAAAAC :REGISTER letmein me@aaronmdjones.net
[2025-01-26 14:07:58] atheme_argon2_compute: argon2_ctx() failed: Memory allocation error
[2025-01-26 14:07:58] crypt_password: ci->crypt() failed for provider 'crypto/argon2'
[2025-01-26 14:07:58] crypt_password: all encryption-capable crypto providers failed
[2025-01-26 14:07:58] <- :00AAAAAAC NOTICE 42XAAAAAC :There was an error setting your password. Please check it for any invalid characters and contact network staff if the issue persists.
```

Screenshot from 2025-01-26 14-08-11

modules/nickserv/sendpass.c
```diff
@@ -198,11 +198,19 @@ ns_cmd_sendpass(struct sourceinfo *si, int parc, char *parv[])
 			else
 				command_success_nodata(si, _("Overriding MARK on the account %s."), entity(mu)->name);
 		}
-		logcommand(si, CMDLOG_ADMIN, "SENDPASS: \2%s\2", name);
 
 		newpass = random_string(16);
```
ilbelkyr: This was already the case before this PR, but fwiw this probably should use `config_options.default_pass_length` instead of hardcoding 16

PR #956, opened 2026-07-20
Add labeled-response support for solanum

skizzerz:
This required a handful of changes in core code as well:

- Actually parse IRCv3 message tags instead of skipping over them; they are stored in a new patricia tree on struct sourceinfo which is NULL if no tags are passed to avoid unnecessary allocation overhead
- New flag for struct ircd to denote that it supports receiving message tags over the server-to-server link; solanum conditionally defines this flag based on the STAG capab, but other protocol modules may unconditionally define it if they know uplinks always support tags
- Move sts to phandler so protocol modules can override it if necessary, allowing a low-level hook point for all outgoing traffic; the existing sts function was renamed to send_line and made non-variadic
- send_line will skip over outgoing tags if the ircd isn't flagged as supporting message tags, allowing modules to safely pass arbitrary tags without needing to worry if they are supported or not (e.g. ChanServ modules could in the future pass +channel-context)
- New handle_command wrapper to dispatch commands, the default impl of which simply calls the handler if it exists
- Advertise the STAG capab in ts6-generic, and set the new message tag flag on the ircd struct if we receive STAG from the uplink
- The solanum module overrides sts to add the server-time tag (required now that we advertise STAG) and the solanum.chat/response tag if one was passed to us
- The solanum module overrides handle_command to send ENCAP ACK at the end of command processing if required by an incoming solanum.chat/response tag

### DIFF
--- a/modules/protocol/solanum.c
+++ b/modules/protocol/solanum.c
@@ -116,7 +116,7 @@ current_time(void)
 	struct tm *tm = NULL;
 	unsigned int ms = 0;
 
-#if defined(TIME_UTC)
+#if defined(HAVE_TIMESPEC_GET)
 	struct timespec ts;
 	timespec_get(&ts, TIME_UTC);
 	tm = gmtime(&ts.tv_sec);
@@ -190,12 +190,12 @@ solanum_sts(const char *fmt, ...)
 void
 solanum_handle_command(struct proto_cmd *pcmd, struct sourceinfo *si, int parc, char *parv[])
 {
-	current_si = si;
-
 	if (pcmd->handler)
+	{
+		current_si = si;
 		pcmd->handler(si, parc, parv);
-
-	current_si = NULL;
+		current_si = NULL;
+	}
 
 	if (si->tags != NULL && ircd->flags & IRCD_MESSAGE_TAGS)
 	{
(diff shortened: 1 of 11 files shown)


modules/protocol/solanum.c
```diff
@@ -107,6 +109,124 @@ solanum_is_valid_hostslash(const char *host)
...
 }
 
+static const char *
+current_time(void)
+{
+	static char buf[32];
+	struct tm *tm = NULL;
+	unsigned int ms = 0;
+
+#if defined(TIME_UTC)
```
aaronmdjones: These need some configure tests to make them more reliable (there is a risk that the macros are defined but the functions are not; this is a C11 function in a C99 codebase); add them to `AC_CHECK_FUNCS` in `m4/atheme-check-build-requirements.m4` which will define `HAVE_TIMESPEC_GET` etc.

modules/protocol/solanum.c
```diff
@@ -107,6 +109,124 @@ solanum_is_valid_hostslash(const char *host)
...
+
+void
+solanum_handle_command(struct proto_cmd *pcmd, struct sourceinfo *si, int parc, char *parv[])
+{
+	current_si = si;
+
+	if (pcmd->handler)
+		pcmd->handler(si, parc, parv);
+
+	current_si = NULL;
```
aaronmdjones:
```C
if (pcmd->handler)
{
    current_si = si;
    pcmd->handler(si, parc, parv);
    current_si = NULL;
}
```

aaronmdjones (changes requested):

aaronmdjones (approved):

